PR #22, opened 2025-05-07
Add error classes

### LINKED ISSUE #1: Implement error types
- JSONParseError
- InvalidRequestError
- MethodNotFoundError
- InvalidParamsError
- InternalError
- TaskNotFoundError
- TastNotCancelableError
- PushNotificationNotSupportedError
- UnsupportedOperationError
- ContentTypeNotSupportedError
- A2AClientError
- A2AClientHttpError
- A2AClientJSONError
- MissingAPIKeyError

e.g., see https://github.com/google/A2A/blob/main/samples/python/common/types.py#L251

kabir:
Fixes #1

### DIFF
--- a/src/main/java/io/a2a/spec/A2AClientError.java
+++ b/src/main/java/io/a2a/spec/A2AClientError.java
@@ -1,4 +1,10 @@
 package io.a2a.spec;
 
 public class A2AClientError extends Exception {
+    public A2AClientError() {
+    }
+
+    public A2AClientError(String message) {
+        super(message);
+    }
 }
--- a/src/main/java/io/a2a/spec/A2AClientError.java
+++ b/src/main/java/io/a2a/spec/A2AClientError.java
@@ -1,0 +1,4 @@
+package io.a2a.spec;
+
+public class A2AClientError extends Exception {
+}
--- a/src/main/java/io/a2a/spec/A2AClientHTTPError.java
+++ b/src/main/java/io/a2a/spec/A2AClientHTTPError.java
@@ -2,16 +2,12 @@ package io.a2a.spec;
 
 import io.a2a.util.Assert;
 
-/**
- * Represents a JSONRPC error.
- */
-public class JSONRPCError {
-
+public class A2AClientHTTPError extends A2AClientError {
     private final int code;
     private final String message;
     private final Object data;
 
-    public JSONRPCError(int code, String message, Object data) {
+    public A2AClientHTTPError(int code, String message, Object data) {
         Assert.checkNotNullParam("code", code);
         Assert.checkNotNullParam("message", message);
         this.code = code;
--- a/src/main/java/io/a2a/spec/A2AClientHTTPError.java
+++ b/src/main/java/io/a2a/spec/A2AClientHTTPError.java
@@ -5,14 +5,12 @@ import io.a2a.util.Assert;
 public class A2AClientHTTPError extends A2AClientError {
     private final int code;
     private final String message;
-    private final Object data;
 
     public A2AClientHTTPError(int code, String message, Object data) {
         Assert.checkNotNullParam("code", code);
         Assert.checkNotNullParam("message", message);
         this.code = code;
         this.message = message;
-        this.data = data;
     }
 
     /**
@@ -31,14 +29,5 @@ public class A2AClientHTTPError extends A2AClientError {
      */
     public String getMessage() {
         return message;
-    }
-
-    /**
-     * Gets the data associated with the error.
-     *
-     * @return the data. May be {@code null}
-     */
-    public Object getData() {
-        return data;
     }
 }
--- a/src/main/java/io/a2a/spec/A2AClientJSONError.java
+++ b/src/main/java/io/a2a/spec/A2AClientJSONError.java
@@ -1,0 +1,4 @@
+package io.a2a.spec;
+
+public class A2AClientJSONError extends A2AClientError {
+}
--- a/src/main/java/io/a2a/spec/InvalidParamsError.java
+++ b/src/main/java/io/a2a/spec/InvalidParamsError.java
@@ -1,0 +1,7 @@
+package io.a2a.spec;
+
+public class InvalidParamsError extends JSONRPCError {
+    public InvalidParamsError(int code, String message, Object data) {
+        super(code, message, data);
+    }
+}
--- a/src/main/java/io/a2a/spec/InvalidParamsError.java
+++ b/src/main/java/io/a2a/spec/InvalidParamsError.java
@@ -1,7 +1,23 @@
 package io.a2a.spec;
 
+import static io.a2a.util.Utils.defaultIfNull;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class InvalidParamsError extends JSONRPCError {
-    public InvalidParamsError(int code, String message, Object data) {
-        super(code, message, data);
+    @JsonCreator
+    public InvalidParamsError(
+            @JsonProperty("code") Integer code,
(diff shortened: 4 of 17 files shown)


src/main/java/io/a2a/spec/A2AClientError.java
```diff
@@ -0,0 +1,4 @@
+package io.a2a.spec;
+
+public class A2AClientError extends Exception {
```
kabir, line 3: Not 100% sure here, the Python implementation is:

```
class A2AClientError(Exception):
    pass

File: src/main/java/io/a2a/spec/A2AClientJSONError.java [hunk 2]
```diff
@@ -0,0 +1,4 @@
+package io.a2a.spec;
+
+public class A2AClientJSONError extends A2AClientError {
```
kabir, line 3: Not totally sure here either, here is the Python

```
class A2AClientJSONError(A2AClientError):
    def __init__(self, message: str):
        self.message = message
        super().__init__(f'JSON Error: {message}')

File: src/main/java/io/a2a/spec/MissingAPIKeyError.java [hunk 3]
```diff
@@ -0,0 +1,5 @@
+package io.a2a.spec;
+
+public class MissingAPIKeyError extends Exception {
```
kabir, line 3: Same doubt as for the A2AClientError

src/main/java/io/a2a/spec/JSONRPCError.java
```diff
@@ -5,10 +5,44 @@
 /**
  * Represents a JSONRPC error.
  */
-public record JSONRPCError(int code, String message, Object data) {
+public class JSONRPCError {
```
fjuma, line 14: Just to check, any reason to convert this to a class? This is immutable and we're using Jackson to convert to/from JSON. It looks like the MCP Java SDK used records for errors like this.

fjuma: Ah, I see now, we need to use a class because the other errors need to extend this one, right?

src/main/java/io/a2a/spec/UnsupportedOperationError.java
```diff
@@ -0,0 +1,8 @@
+package io.a2a.spec;
+
+public class UnsupportedOperationError extends JSONRPCError {
+
+    public UnsupportedOperationError(int code, String message, Object data) {
```
fjuma: For classes like this that extend `JSONRPCError`, I believe we'll need the `@JsonCreator` annotation on the constructor and `@JsonProperty("blah")` annotations on each parameter to be able to successfully create the object from JSON. Let me know if I'm missing something though.

src/main/java/io/a2a/spec/UnsupportedOperationError.java
```diff
@@ -0,0 +1,8 @@
...
+        super(code, message, data);
```
fjuma: Looks like the spec specifies default values for code and message:

https://github.com/google/A2A/blob/main/specification/json/a2a.json#L1450
https://github.com/google/A2A/blob/main/samples/python/common/types.py#L303-L304

Looks like the other error types also have defaults that we should use as well.

src/main/java/io/a2a/spec/A2AClientError.java
fjuma: Maybe we could add a constructor here that just takes a string message?

src/main/java/io/a2a/spec/A2AClientHTTPError.java
```diff
@@ -0,0 +1,44 @@
...
+    private final int code;
+    private final String message;
+    private final Object data;
+
+    public A2AClientHTTPError(int code, String message, Object data) {
+        Assert.checkNotNullParam("code", code);
+        Assert.checkNotNullParam("message", message);
+        this.code = code;
+        this.message = message;
+        this.data = data;
```
fjuma: Is data needed here? I don't see it in the python implementation for this one.

src/main/java/io/a2a/spec/A2AClientJSONError.java
fjuma: Looks like they're just prefixing the message string with "JSON Error". We can always do that later if we think it's useful. I see now that A2AClientJSONError isn't in a2a.json so what we do here is mostly up to us.

src/main/java/io/a2a/spec/JSONRPCError.java
kabir: @fjuma Yeah, I wasn't able to inherit from Records. It says they are final

src/main/java/io/a2a/spec/UnsupportedOperationError.java
kabir: I need to read up a bit on this :-)

src/main/java/io/a2a/spec/A2AClientError.java
maeste: pass means "no implementation" in python. It's like an interface in Java, just for typing

maeste: well bad explanation...since it extends Exception class, in python it inherits also its constructor. IOW pass just pass anything to the super class. So this code should work

```
def a_method():
    raise A2AClientError("Connection failed")

try:
    amehod()
except A2AClientError as e:
    print(f"Exception message: {e}")
```
    
 and it should print something like: Exception message: Connection failed

kabir:
Hi @fjuma PR updated. 

For the root JSONRPCError class I wasn't sure if this should have the annotations, or if it is always one of the subclasses. I originally made it abstract, but looking at spec json, a lot of the Response classes simply have the error as JSONRPCError. So, I think the errors must be getting deserialized on the client, and thus ended up putting the annotations on this class anyway.

I see in the test it is simply deserialized as JSONRPCError. So I'm now wondering a bit if the sub-classes actually need the annotations, but don't know yet whether they are needed or not. Perhaps they are used for the serialization.

src/main/java/io/a2a/spec/InvalidParamsError.java
```diff
@@ -0,0 +1,23 @@
...
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class InvalidParamsError extends JSONRPCError {
+    @JsonCreator
+    public InvalidParamsError(
+            @JsonProperty("code") Integer code,
+            @JsonProperty("message") String message,
+            @JsonProperty("data") Object data) {
+        super(
+                defaultIfNull(code, -32702),
```
fjuma, line 19: This should be -32602

src/main/java/io/a2a/spec/UnsupportedOperationError.java
```diff
@@ -0,0 +1,23 @@
...
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class UnsupportedOperationError extends JSONRPCError {
+    @JsonCreator
+    public UnsupportedOperationError(
+            @JsonProperty("code") Integer code,
+            @JsonProperty("message") String message,
+            @JsonProperty("data") Object data) {
+        super(
+                defaultIfNull(code, -32704),
```
fjuma, line 19: This should be -32004

fjuma:
Looks good, just a couple small comments on the error codes.

> I see in the test it is simply deserialized as JSONRPCError. So I'm now wondering a bit if the sub-classes actually need the annotations, but don't know yet whether they are needed or not. Perhaps they are used for the serialization.

For `Part` and its sub-classes, I was able to use `@JsonTypeInfo` and `@JsonSubTypes` to make sure the correct class gets instantiated:

https://github.com/fjuma/a2a-java-sdk/blob/main/src/main/java/io/a2a/spec/Part.java#L12-L21

For `JSONRPCError` and its sub-classes, there isn't a `type` field so this approach wouldn't work there to get the correct class to be instantiated. Maybe we'd need a custom deserializer for something like that? We could always do that as a follow-up issue.

fjuma:
@kabir I created https://github.com/fjuma/a2a-java-sdk/issues/23 to track looking into deserialization for classes that extend JSONRPCError to see if the correct class can be instantiated based on the error code.

I'll go ahead and merge this PR and submit a small fix for the two error codes. Thanks!

PR #27, opened 2025-05-08
Implement A2AClient getTask, cancelTask, and getTaskPushNotificationConfig

fjuma:
Closes: 

* https://github.com/fjuma/a2a-java-sdk/issues/14
* https://github.com/fjuma/a2a-java-sdk/issues/15
* https://github.com/fjuma/a2a-java-sdk/issues/16

### DIFF
--- a/src/main/java/io/a2a/spec/CancelTaskRequest.java
+++ b/src/main/java/io/a2a/spec/CancelTaskRequest.java
@@ -37,4 +37,35 @@ public final class CancelTaskRequest extends JSONRPCRequest {
         this.method = method;
         this.params = paramsMap;
     }
+
+    public static class Builder {
+        private String jsonrpc;
+        private Object id;
+        private String method;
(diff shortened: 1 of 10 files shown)


src/main/java/io/a2a/spec/CancelTaskRequest.java
```diff
@@ -37,4 +37,35 @@ public CancelTaskRequest(@JsonProperty("jsonrpc") String jsonrpc, @JsonProperty(
         this.method = method;
         this.params = paramsMap;
     }
+
+    public static class Builder {
+        private String jsonrpc;
```
kabir, line 42: Not sure if this is handled somewhere, but it says that the default for this is "2.0". Mentioning this since I am trying to deal with the same on the server side

fjuma: Good point, will address that separately.

PR #34, opened 2025-05-13
Initial bunch of spec related updates

fjuma:
I've started making updates as per https://github.com/google/A2A/pull/415.

I'm going to create separate issues to track the remaining updates that need to be made.


src/main/java/io/a2a/spec/JSONRPCRequest.java
```diff
@@ -16,9 +16,9 @@
  */
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public sealed class JSONRPCRequest implements JSONRPCMessage permits SendTaskRequest, SendTaskStreamingRequest,
+public sealed class JSONRPCRequest implements JSONRPCMessage permits SendTaskStreamingRequest,
         GetTaskRequest, CancelTaskRequest, SetTaskPushNotificationRequest, GetTaskPushNotificationRequest,
-        TaskResubscriptionRequest {
```
maeste, line 21: does it make sense to be non-abstract?
Can we have a JSONRPCRequest which is none of the permitted subclasses?

src/main/java/io/a2a/spec/JSONRPCResponse.java
```diff
@@ -13,8 +13,9 @@
  */
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public sealed class JSONRPCResponse implements JSONRPCMessage permits SendTaskResponse, SendTaskStreamingResponse,
-        GetTaskResponse, CancelTaskResponse, SetTaskPushNotificationResponse, GetTaskPushNotificationResponse {
```
maeste, line 17: does it make sense to be non-abstract?
Can we have a JSONRPCResponse which is none of the permitted subclasses?

src/main/java/io/a2a/spec/JSONRPCRequest.java
fjuma: I haven't come across usages yet where a JSONRPCRequest isn't one of the permitted subclasses. I'll make it abstract.

src/main/java/io/a2a/spec/JSONRPCResponse.java
fjuma: I'll do the same here.

src/test/java/io/a2a/client/A2AClientTest.java
```diff
@@ -73,37 +71,46 @@ public void tearDown() {
...
-        Message message = toUserMessage("tell me a joke");
-        TaskSendParams params = new TaskSendParams.Builder()
-                .id("task-1234")
-                .sessionId("session-1234")
+        Message message = new Message.Builder()
+                .role(Message.Role.USER)
+                .parts(Collections.singletonList(new TextPart("tell me a joke")))
+                .contextId("context-1234")
+                .messageId("message-1234")
+                .build();
```
maeste, line 94: This is a general comment, not impacting this PR...but I'm wondering who ensure contextId and messageID are unique

fjuma: Good question. In general, a messageId is determined by the creator of the message (client or server). If it's not explicitly specified, it's randomly generated. The contextId will be created by the server. I'll create an issue to track this.

PR #58, opened 2025-05-19
Add the event classes from the Python implementation

kabir:


src/main/java/io/a2a/server/events/EventConsumer.java
```diff
@@ -0,0 +1,84 @@
...
+                    boolean isFinalEvent = false;
+                    if (event instanceof TaskStatusUpdateEvent tue && tue.isFinal()) {
+                        isFinalEvent = true;
+                    } else if (event instanceof Message) {
+                        isFinalEvent = true;
+                    } else if (event instanceof Task task) {
+                        switch (task.getStatus().state()) {
+                            case COMPLETED:
+                            case CANCELED:
+                            case FAILED:
```
fjuma, line 69: I think we need to add REJECTED and UNKNOWN here too:

https://github.com/google/a2a-python/blob/main/src/a2a/server/events/event_consumer.py#L79-L83

kabir: @fjuma I have added that locally since today's changes have that. The above set was in yesterday's updates ;-)

fjuma: Ah, ok, great :)

PR #64, opened 2025-05-22
example of java client running against python server

maeste:
It doesn't work, and we need to discuss about this


fjuma:
I've included this commit here along with a note to temporarily use the v0.2.1a1 version of the server with the example until the spec changes have been applied:

https://github.com/fjuma/a2a-java-sdk/pull/66

PR #70, opened 2025-05-26
Add initial A2AServerResource and tests

fjuma:


src/main/java/io/a2a/spec/JSONRPCErrorResponse.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * A JSON RPC error response.
+ */
+@JsonInclude(JsonInclude.Include.NON_ABSENT)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public final class JSONRPCErrorResponse extends JSONRPCResponse<Void> {
+
+    @JsonCreator
+    public JSONRPCErrorResponse(@JsonProperty("jsonrpc") String jsonrpc, @JsonProperty("id") Object id,
+                                @JsonProperty("result") Void result, @JsonProperty("error") JSONRPCError error) {
```
kabir, line 17: Is the result needed here?

fjuma: If an error occurs, the result in the JSON RPC response is always null.

PR #86, opened 2025-06-02
Add majority of new JSONRPCHandler tests

kabir:


src/main/java/io/a2a/server/requesthandlers/JSONRPCHandler.java
```diff
@@ -53,14 +55,25 @@ public SendMessageResponse onMessageSend(SendMessageRequest request) {
...
     public Flow.Publisher<SendStreamingMessageResponse> onMessageSendStream(SendStreamingMessageRequest request) {
-        Flow.Publisher<StreamingEventKind> publisher = requestHandler.onMessageSendStream(request.getParams());
-        return convertingProcessor(publisher, event -> {
-            try {
-                return new SendStreamingMessageResponse(request.getId(), event);
-            } catch (JSONRPCError error) {
-                return new SendStreamingMessageResponse(request.getId(), error);
-            }
-        });
+        if (!agentCard.capabilities().streaming()) {
```
fjuma, line 60: We should probably have a similar check for the task resubscription case, right?

kabir: Yeah, I probably missed it due to a lacking test OR one the remaining tests testing that :-)

kabir: @fjuma actually this is a bit different, and there is no existing test.

The signature is `Flow.Publisher<SendStreamingMessageResponse> onResubscribeToTask`, so the results are streamed. If the server does not support streaming, it seems we cannot return anything.

So perhaps this needs to be handled at the HTTP level? e.g with a 404.

Then again, perhaps it doesn't make sense to resubscribe to something they were not able to do in the first place (ie send a streaming request).

kabir:
@fjuma I thought again, and have added the missing test and check

PR #95, opened 2025-06-05
Fire and forget task execution

maeste:
FireAndForgetAgentExecutor


kabir:
@maeste The agent execution already happens in a background thread. Please see DefaultRequestHandler.registerAndExecuteAgentAsync().  Then the reading of the results from the queue happens on the calling thread.

That said, perhaps we could improve the executor I have to be more in line with yours. And also perhaps find a way to handle cancelling better in the main code (if possible)

PR #108, opened 2025-06-10
Switch out OkHttpClient for HttpClient from the JDK

### LINKED ISSUE #59: Implement InMemoryPushNotifier.sendNotification() so it sends notification
The class uses an http client, which will also need implementing

kabir:
Fixes #59


tck/src/main/java/io/a2a/examples/helloworld/server/A2AHttpClientProducer.java
```diff
@@ -0,0 +1,17 @@
+package io.a2a.examples.helloworld.server;
```
fjuma: We should also make this change in the examples directory.

kabir: Updated

PR #109, opened 2025-06-11
Cleanup the AgentExecutors and use TaskUpdater where possible

kabir:


tck/src/main/java/io/a2a/tck/server/AgentExecutorProducer.java
```diff
@@ -42,22 +43,12 @@ public void execute(RequestContext context, EventQueue eventQueue) throws JSONRP
...
             System.out.println("====> task set to WORKING, starting background execution");
 
-//            // Fire and forget - start the task but don't wait for it
-//            CompletableFuture<Void> taskFuture = CompletableFuture
-//                .runAsync(() -> executeTaskInBackground(context, eventQueue), taskExecutor);
-            
-//            // Store the future for potential cancellation
-//            runningTasks.put(context.getTaskId(), taskFuture);
-            
             // Method returns immediately - task continues in background
```
fjuma, line 52: There's a executeTaskInBackground method in this class but it looks to be unused?

PR #141, opened 2025-06-30
[DISCUSSION ONLY - DO NOT MERGE] Added CheckStyle and Spotless plugins for code checking and formatting

sunyuhan1998:
We need to add a code style checking and automatic formatting plugin to unify the code style and automatically handle common formatting issues. 

In this PR, the `Checkstyle` and `Spotless` plugins have been introduced. For `Checkstyle`, only some of the previously discussed rules have been included in the rule file `src/checkstyle/checkstyle.xml`, such as disallowing importing unused classes and disallowing wildcard imports. More rules can be added in the future if needed. 

As for `Spotless`, it is currently using the default rules. Both plugins run during the Maven `validate` phase. 

You can directly run `mvn clean validate` to trigger the checks from both plugins, and run `mvn spotless:apply` to trigger Spotless to automatically format the code.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,6 +29,8 @@
         <quarkus.platform.version>3.22.3</quarkus.platform.version>
         <rest-assured.version>5.5.1</rest-assured.version>
         <slf4j.version>2.0.17</slf4j.version>
+        <maven-checkstyle-plugin.version>3.5.0</maven-checkstyle-plugin.version>
+        <puppycrawl-tools-checkstyle.version>9.3</puppycrawl-tools-checkstyle.version>
 
         <!-- Redirect test output to file -->
         <maven.test.redirectTestOutputToFile>true</maven.test.redirectTestOutputToFile>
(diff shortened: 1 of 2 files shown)


kabir:
@sunyuhan1998 Just about to get on a plane, but will look later/tomorrow. 

In the meantime, it could be good to see what kind of stuff it reports against the existing codebase. Maybe as a link to a gist, if it is very long :-)

kabir:
BTW I converted the PR to a draft, so that it cannot be merged

sunyuhan1998:
> @sunyuhan1998 Just about to get on a plane, but will look later/tomorrow.
> 
> In the meantime, it could be good to see what kind of stuff it reports against the existing codebase. Maybe as a link to a gist, if it is very long :-)

Sure, no problem. This is just a rough experimental attempt. I think it will still take some time before we actually integrate it, and during that period we can further discuss and refine the plan in detail.

PR #161, opened 2025-07-04
fix: Fixed the intermittent failure issue in `testOnMessageStreamNewMessageSendPushNotificationSuccess`

### LINKED ISSUE #140: [Bug]: Intermittent failures in JSONRPCHandlerTest.testOnMessageStreamNewMessageSendPushNotificationSuccess()
### What happened?

The test fails quite often when running it. Seems to be some kind of race condition, which I have not had time to dig into yet.

### Relevant log output

```shell

```

### Code of Conduct

sunyuhan1998:
The reason for its intermittent failure, I believe, is as follows:

We initially set up the `CountDownLatch` with a count of 6, and we will call `latch.countDown()` at the following three locations:

- A. in `onSubscribe`
- B. in the `finally` block of the `post` method in `HttpClient`
- C. in `onNext`

Ideally, when the test method runs, the sequence of `countDown()` calls should be like this:
**A B C B C B C**

The `countDown()` function was called a total of 7 times!

This explains why we sometimes see only **2 elements** in `results` when we do `assertEquals(3, results.size())`. It's because the last `onNext` may not have completed execution yet.

**What this PR does:**  

We could fix this by initializing the `CountDownLatch` to 7. However, considering the purpose of this test method, I chose instead to keep the count at 6 and **remove the `latch.countDown()` call in `onSubscribe`**.

Fixes #140

### DIFF
--- a/sdk-server-common/src/test/java/io/a2a/server/requesthandlers/JSONRPCHandlerTest.java
+++ b/sdk-server-common/src/test/java/io/a2a/server/requesthandlers/JSONRPCHandlerTest.java
@@ -700,7 +700,6 @@ public class JSONRPCHandlerTest {
                 public void onSubscribe(Flow.Subscription subscription) {
                     subscriptionRef.set(subscription);
                     subscription.request(1);
-                    latch.countDown();
                 }
 
                 @Override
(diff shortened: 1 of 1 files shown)


sunyuhan1998:
I'm not entirely sure if the root cause I identified is accurate, but for now the unit test has passed. Could you please help review it @kabir ? Thank you!

PR #162, opened 2025-07-04
feat: Add a2a-tck runs to CI for PRs (GH-152)

### LINKED ISSUE #152: Add a2a-tck runs to CI for PRs
Currently, when a PR is submitted, only the tests in the a2a-java project get run:

https://github.com/a2aproject/a2a-java/blob/main/.github/workflows/build-and-test.yml

We should also run the a2a-tck to avoid regressions:

https://github.com/maeste/a2a-tck

We can use the 0.2.3 a2a-tck tag for now.

To run the TCK:
* We need to first start the Java system under test by going to the `a2a-java/tck` directory and running `mvn quarkus:dev`.
* Then, the TCK can be run using the instructions from its README (i.e., `./run_tck.py --sut-url http://localhost:9999 --category all --compliance-report report.json`)

When we apply future spec updates, we can update the GitHub Action accordingly to checkout a newer a2a-tck spec tag.

dev-jonghoonpark:
# Description

Added the [a2a-tck](https://github.com/maeste/a2a-tck) test to the GitHub Actions workflow of the `a2a-java` project.

Tested locally using the following commands.

```
act -P ubuntu-latest=quay.io/jamezp/act-maven -W '.github/workflows/tck-test.yml' 
```

Fixes #152 🦕

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -21,5 +21,5 @@ jobs:
         java-version: '17'
         distribution: 'temurin'
         cache: maven
-    - name: Build with Maven
+    - name: Build with Maven and run tests
       run: mvn -B package --file pom.xml -fae
--- a/.github/workflows/run-tck.yml
+++ b/.github/workflows/run-tck.yml
@@ -12,7 +12,7 @@ on:
 env:
   # Tag of the TCK
   TCK_VERSION: 0.2.3
-  # Tells astral-sh/setup-uv@v5 to not need a venv, and instead use system
+  # Tells uv to not need a venv, and instead use system
   UV_SYSTEM_PYTHON: 1
 
 # Only run the latest job
@@ -43,18 +43,19 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
-      - name: Install uv
+          python-version-file: "tck/a2a-tck/pyproject.toml"
+      - name: Install uv and Python dependencies
         run: |
           pip install uv
           uv pip install -e .
         working-directory: tck/a2a-tck
       - name: Build with Maven, skipping tests
         run: mvn -B install -DskipTests
-      - name: Start SUT & Run TCK
+      - name: Start SUT
+        run: mvn -B quarkus:dev &
+        working-directory: tck
+      - name: Wait for SUT to start
         run: |
-          mvn -B quarkus:dev &
-          
           URL="http://localhost:9999/.well-known/agent.json"
           EXPECTED_STATUS=200
           TIMEOUT=120
@@ -87,7 +88,7 @@ jobs:
             sleep "$RETRY_INTERVAL"
           done
           
-          # Run TCK
-          cd a2a-tck
+      - name: Run TCK
+        run: |
           ./run_tck.py --sut-url http://localhost:9999 --category all --compliance-report report.json
-        working-directory: tck
+        working-directory: tck/a2a-tck
--- a/.github/workflows/run-tck.yml
+++ b/.github/workflows/run-tck.yml
@@ -1,0 +1,93 @@
+name: Run TCK
+
+on:
+  # Handle all branches for now
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+env:
+  # Tag of the TCK
+  TCK_VERSION: 0.2.3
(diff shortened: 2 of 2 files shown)


.github/workflows/tck-test.yml
```diff
@@ -0,0 +1,64 @@
...
+    
+          # Check if loop exited normally (SUT is not up)
+          if ! curl -s http://localhost:9000/q/health | grep '"status": "UP"' > /dev/null; then
+            echo "❌ SUT failed to start within timeout."
+            echo "Full quarkus.log for debugging:"
+            cat ./quarkus.log
+            exit 1
+          fi
+    
+          git clone --depth 1 --branch 0.2.3 https://github.com/maeste/a2a-tck.git
```
kabir: Please put 0.2.3 in an environment variable at the top level of the workflow. It will make it easier to change it when we go for the next version

.github/workflows/tck-test.yml
```diff
@@ -0,0 +1,64 @@
+name: Tck Test
+
+on:
+  # Handle all branches for now
+  push:
```
kabir: We think we should just target the main branch. We might e.g start a feature branch in the main repo to bring in changes to the next specification version. And then those will fail.

.github/workflows/tck-test.yml
```diff
@@ -0,0 +1,64 @@
...
+  pull_request:
```
kabir: Please target the main branch

.github/workflows/tck-test.yml
```diff
@@ -0,0 +1,64 @@
...
+          pip install uv
+      - name: Maven Install
+        run: mvn clean install -DskipTests=true
+      - name: Start SUT and run test
+        run: |
+          cd ./tck
+          mvn clean install -DskipTests=true
+          mvn -B quarkus:run > quarkus.log 2>&1 &
+          for i in {1..60}; do
+            if curl -s http://localhost:9000/q/health | grep '"status": "UP"' > /dev/null; then
```
kabir: Instead of the health endpoint, we can monitor the agent card url `http://localhost:9999/.well-known/agent.json` which should return status code 200, e.g:

```
        URL="http://localhost:9999/.well-known/agent.json"
        HTTP_STATUS=$(curl --output /dev/null --silent --write-out "%{http_code}" "$URL") || true
```

Then if HTTP_STATUS is 200 we're good, otherwise we should sleep a bit and try again.

tck/pom.xml
```diff
@@ -35,6 +35,10 @@
             <groupId>jakarta.ws.rs</groupId>
             <artifactId>jakarta.ws.rs-api</artifactId>
         </dependency>
+        <dependency>
```
kabir: Please remove this dependency since we will use the agent card to check this instead

kabir (changes requested):
Hi @dev-jonghoonpark, thanks for the contribution!

I actually also opened https://github.com/a2aproject/a2a-java/pull/163 for this, but we can use yours since yours came first. So I have closed mine.

A few small suggestions thanks!

dev-jonghoonpark:
@kabir 
Thank you for your feedback.
I’ve made the changes based on your comments and your PR, and pushed the updates.

---

While working on it, I also wanted to separate the 'Wait for server to start' step and the 'Run TCK' step like you did.
However, when I separated the steps, the server would shut down before the TCK run step.
So I ended up combining them into a single step.

Did you happen to encounter any issues with this when you tested it?

kabir:
No, but then I just tried it a few times (i had a very hard time with the wait for server step, until I googled the `|| true` trick).

We can look more on Monday?

dev-jonghoonpark:
Certainly. Thank you. : )

kabir:
Hi @dev-jonghoonpark since I am a project admin, I took the liberty to rebase your PR on main, and add a commit which splits starting Quarkus, waiting for it to start and running the TCK into separate steps.

Hope you don't mind :-) 

I think we will merge it as soon as possible. If you don't see this before it is merged, and you have some concerns, please raise those as a follow-up PR ?

dev-jonghoonpark:
@kabir 

Of course, no worries!
I agree with your approach, and I also think that direction makes much more sense as long as there are no issues.

If everything worked fine in your testing process,
then the issue might be specific to [act](https://github.com/nektos/act), which I used for local testing.

Thank you :)
Hope you have a great day!

kabir:
Ah ok, I've not tried act in years. I just try out these things in my private repo, and it works like this there

PR #219, opened 2025-08-08
feat: Refactor GrpcHandler to not use CDI and be more a utility

kabir:


transport/grpc/src/main/java/io/a2a/grpc/handler/GrpcHandler.java
```diff
@@ -46,29 +43,14 @@
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 
-import java.util.HashMap;
-import java.util.Map;
-
-@ApplicationScoped
```
fjuma, line 52: Does this mean the following line can now be removed from application.properties too?
```
# The GrpcHandler @ApplicationScoped annotation is not compatible with Quarkus
quarkus.arc.exclude-types=io.a2a.grpc.handler.GrpcHandler
```

fjuma (approved):

PR #223, opened 2025-08-12
chore: Adding some testing on toProto conversion.

ehsavoie:
Checking that optional parameters not being set doesn't lead to a NullPointerException.

### DIFF
--- a/spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
+++ b/spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
@@ -1,5 +1,6 @@
 package io.a2a.grpc.utils;
 
+
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDateTime;
@@ -61,6 +62,7 @@ import io.a2a.spec.TextPart;
  * Utility class to convert between GRPC and Spec objects.
  */
 public class ProtoUtils {
+
     public static class ToProto {
 
         public static io.a2a.grpc.AgentCard agentCard(AgentCard agentCard) {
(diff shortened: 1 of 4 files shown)


spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
```diff
@@ -143,8 +145,8 @@ public static io.a2a.grpc.Task task(Task task) {
         public static io.a2a.grpc.Message message(Message message) {
             io.a2a.grpc.Message.Builder builder = io.a2a.grpc.Message.newBuilder();
             builder.setMessageId(message.getMessageId());
-            builder.setContextId(message.getContextId());
-            builder.setTaskId(message.getTaskId());
+            builder.setContextId(requireNonNullElse(message.getContextId(), ""));
```
fjuma: Just wondering, instead of setting the default value ourselves, should we only call setContextId if non null and then let the grpc builder set the default for us?

ehsavoie: Yes, makes sense :+1:

spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
```diff
@@ -182,8 +188,8 @@ public static io.a2a.grpc.TaskArtifactUpdateEvent taskArtifactUpdateEvent(TaskAr
             builder.setTaskId(event.getTaskId());
             builder.setContextId(event.getContextId());
             builder.setArtifact(artifact(event.getArtifact()));
-            builder.setAppend(event.isAppend() == null ? false : event.isAppend());
-            builder.setLastChunk(event.isLastChunk() == null ? false : event.isLastChunk());
+            builder.setAppend(requireNonNullElse(event.isAppend(), false));
```
fjuma: Same thing here and below, I think we can just call the method if non null and let the builder set the right default.

spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
```diff
@@ -143,8 +145,12 @@ public static io.a2a.grpc.Task task(Task task) {
...
             io.a2a.grpc.Message.Builder builder = io.a2a.grpc.Message.newBuilder();
             builder.setMessageId(message.getMessageId());
-            builder.setContextId(message.getContextId());
-            builder.setTaskId(message.getTaskId());
+            if (message.getContextId() != null) {
+                builder.setContextId(message.getContextId());
+            }
+            if (message.getTaskId() != null) {
+                builder.setTaskId(message.getTaskId());
+            }
```
fjuma, line 153: I think the corresponding `message` method in `FromProto` will need to be updated so the value is only set if non empty.

PR #227, opened 2025-08-13
fix: issue 226

### LINKED ISSUE #226: [Bug]:
### What happened?

A TaskStatus update message is optional. ProtoUtils.FromProto doesn't take that into account, trying to convert the update message in every stituation

### Relevant log output

```shell

```

### Code of Conduct

ehsavoie:
Checking if the TaskUpdate has an update message before trying to convert it.

Fixes #226 🦕

### DIFF
--- a/spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
+++ b/spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
@@ -3,8 +3,8 @@ package io.a2a.grpc.utils;
 
 import java.nio.charset.StandardCharsets;
 import java.time.Instant;
-import java.time.LocalDateTime;
 import java.time.ZoneOffset;
+import java.time.OffsetDateTime;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java
```diff
@@ -822,9 +822,12 @@ private static DataPart dataPart(io.a2a.grpc.DataPart dataPart) {
         }
 
         private static TaskStatus taskStatus(io.a2a.grpc.TaskStatus taskStatus) {
-            return new TaskStatus(
-                    taskState(taskStatus.getState()),
-                    message(taskStatus.getUpdate()),
+            TaskState state = taskState(taskStatus.getState());
```
fjuma, line 892: Can taskStatus.getState() return null? I don't seem to see that here:

https://github.com/a2aproject/a2a-java/blob/main/spec-grpc/src/main/java/io/a2a/grpc/TaskStatus.java#L75

Let me know if I'm missing something.

fjuma: @ehsavoie Just bumping this question again in case you missed it.

PR #238, opened 2025-08-25
feat: Provide a more friendly builder-based way to configure transpor…

brasseld:
…t configuration

This PR is a proposal to make the transport configuration specific to a given transport implementation.

This one will ease developer experience by providing a builder for configuring a Client in a more stricter way compare to what exists today.

Also, this will help at getting rid of strong coupling with different transport implementations.
Two benefits:

Get rid of dependencies which may not be required (grpc dependencies if you only go for jsonrpc)
Easier to provide new transport implementation (such as HTTP+JSON).

# Description

- [x] Follow the [`CONTRIBUTING` Guide](../CONTRIBUTING.md).
- [x] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [x] Ensure the tests pass


client/base/src/test/java/io/a2a/client/ClientBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+        Assertions.assertTrue(exception.getMessage().contains("No compatible transport found"));
+    }
+
+    @Test
+    public void shouldCreateJSONRPCClient() throws A2AClientException {
+        Client client = Client
+                .from(card)
+                .useClientPreference(true)
+                .withTransport(JSONRPCTransport.class, new JSONRPCTransportConfigBuilder()
+                        .addInterceptor(null)
```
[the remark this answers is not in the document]
brasseld: Expected: that is for the purpose of the test.

README.md
```diff
@@ -243,21 +243,14 @@ that you can use to create your A2A `Client`.
 
 ### 2. Add one or more dependencies on the A2A Java SDK Client Transport(s) you'd like to use
 
-You need to add a dependency on at least one of the following client transport modules:
+By default, the sdk-client is coming with the JSONRPC transport.
```
fjuma: Just wondering about this part. In https://github.com/a2aproject/A2A/pull/896, it was discussed that JSON-RPC isn't a required transport for an A2A server and the spec was updated to indicate that all three transports are considered equal in status and a server might choose to only support a custom transport and in that case the client might only support a custom transport too.

README.md
```diff
@@ -305,14 +298,18 @@ Consumer<Throwable> errorHandler = error -> {
     ...
 };
 
-// Create the client using ClientFactory
-ClientFactory clientFactory = new ClientFactory(clientConfig);
-Client client = clientFactory.create(agentCard, consumers, errorHandler);
+// Create the client using the builder
+Client client = Client
+        .from(finalAgentCard)
```
fjuma: Instead of `Client.from`, just wondering if using the `Builder` directly might be more intuitive for the user, e.g.,

```
new Client.Builder()
    .agentCard(...)
    .clientConfig(...)
    ....
```

README.md
```diff
@@ -339,20 +339,19 @@ ClientConfig clientConfig = new ClientConfig.Builder()
...
-Function<String, Channel> channelFactory = agentUrl -> {
-    return ManagedChannelBuilder.forTarget(agentUrl)
-            ...
-            .build();
-};
-
-// Create gRPC transport configuration
-GrpcTransportConfig grpcConfig = new GrpcTransportConfig(channelFactory);
+// Create a channel from agent URL
+Channel channel = ManagedChannelBuilder.forTarget(agentUrl).build();
```
fjuma: What about the case where the `agentUrl` for gRPC isn't known when creating the client? For example, the agent card contains the additional URLs and transports supported by the agent and gRPC could be found there. During the transport selection process, if gRPC is selected, we'll have the agentUrl at that point. The channel factory approach allows the channel to be created using that URL. I believe the Python SDK also does something similar for this.

README.md
```diff
@@ -324,13 +321,16 @@ as follows:
...
+// Configure the client settings
 ClientConfig clientConfig = new ClientConfig.Builder()
         .setAcceptedOutputModes(List.of("text"))
-        .setClientTransportConfigs(List.of(jsonrpcConfig))
+        .build();
+
+Client client = Client
+        .from(agentCard)
+        .clientConfig(clientConfig)
+        .withJsonRpcTransport(new JSONRPCTransportConfigBuilder()
```
fjuma: Looks like a transport-specific method is only being used for JSON-RPC? I have the same concern here as above - all three transports should be considered equal in status.

client/transport/spi/src/main/java/io/a2a/client/transport/spi/ClientTransportConfigBuilder.java
```diff
@@ -0,0 +1,26 @@
+package io.a2a.client.transport.spi;
+
+import io.a2a.client.transport.spi.interceptors.ClientCallInterceptor;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author David BRASSELY (david.brassely at graviteesource.com)
```
fjuma: Just a very minor nit. We haven't been including author tags in this repo so might be good to remove this here and in other files for consistency.

client/base/pom.xml
```diff
@@ -20,15 +20,24 @@
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>a2a-java-sdk-client-config</artifactId>
+            <artifactId>a2a-java-sdk-http-client</artifactId>
+            <version>${project.version}</version>
```
fjuma: The project.versions can be removed here as long as they are specified in the dependency management section in the parent pom.

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+
+    public ClientBuilder withJsonRpcTransport(JSONRPCTransportConfig config) {
+        return withTransport(JSONRPCTransport.class, config);
+    }
+
+    public ClientBuilder withJsonRpcTransport() {
+        return withTransport(JSONRPCTransport.class, new JSONRPCTransportConfigBuilder());
+    }
+
+    public ClientBuilder addStreamConsumer(BiConsumer<ClientEvent, AgentCard> consumer) {
```
fjuma: The consumers are used for both the streaming and non-streaming case. Would be good to rename this to `addConsumer` or `addClientEventConsumer`

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+        this.consumers.add(consumer);
+        return this;
+    }
+
+    public ClientBuilder addStreamConsumers(List<BiConsumer<ClientEvent, AgentCard>> consumers) {
```
fjuma: Same here.

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+        this.consumers.addAll(consumers);
+        return this;
+    }
+
+    public ClientBuilder streamErrorHandler(Consumer<Throwable> streamErrorHandler) {
```
fjuma: Might be good to rename to `addStreamingErrorHandler`

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+
+        return this;
+    }
+
+    public ClientBuilder useClientPreference(boolean useClientPreference) {
+        this.useClientPreference = useClientPreference;
+        return this;
+    }
+
+    public ClientBuilder withJsonRpcTransport(JSONRPCTransportConfigBuilder configBuilder) {
```
fjuma: Maybe we should use `add` instead of `with` for the transports to make it clear that more than one can be configured.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportProvider.java
```diff
@@ -1,41 +1,37 @@
...
-                    Channel channel = grpcTransportConfig.getChannelFactory().apply(agentUrl);
-                    return new GrpcTransport(channel, agentCard);
-                }
-            }
+
+        Channel channel = grpcTransportConfig.getChannel();
+
+        // no channel factory configured
+        if (channel == null) {
+            channel = ManagedChannelBuilder.forTarget(agentUrl).build();
```
fjuma: One of the cons of creating the channel ourselves is that we're responsible for closing it. Requiring the channel factory to be configured allows the user to be responsible for the channel.

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransportConfig.java
```diff
@@ -1,12 +1,16 @@
...
+import io.a2a.client.transport.spi.ClientTransportConfig;
 import io.a2a.client.http.A2AHttpClient;
 
-public class JSONRPCTransportConfig implements ClientTransportConfig {
+public class JSONRPCTransportConfig extends ClientTransportConfig<JSONRPCTransport> {
 
     private final A2AHttpClient httpClient;
 
+    public JSONRPCTransportConfig() {
+        this.httpClient = null;
```
fjuma, line 11: Shouldn't we default to new JdkA2AHttpClient() here?

tests/server-common/src/test/java/io/a2a/server/apps/common/AbstractA2AServerTest.java
```diff
@@ -137,8 +136,9 @@ protected AbstractA2AServerTest(int serverPort) {
     /**
      * Get the transport configs to use for this test.
      */
-    protected List<ClientTransportConfig> getClientTransportConfigs() {
-        return new ArrayList<>();
+    protected void configureTransport(ClientBuilder builder) {
+        // Include jsonRPC transport by default
+        builder.withJsonRpcTransport();
```
fjuma: I don't think we should include this by default.

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+        // Retrieve the configuration associated to the preferred transport
+        ClientTransportConfig<? extends ClientTransport> clientTransportConfig = clientTransports.get(transportProtocolClass);
+
+        return clientTransportProvider.create(clientTransportConfig, agentCard, agentInterface.url());
+    }
+
+    private Map<String, String> getServerPreferredTransports() {
+        Map<String, String> serverPreferredTransports = new LinkedHashMap<>();
+        serverPreferredTransports.put(agentCard.preferredTransport(), agentCard.url());
+        for (AgentInterface agentInterface : agentCard.additionalInterfaces()) {
```
fjuma: I think this is missing some of the updates made to the corresponding methods from the previous ClientFactory. For example, I believe I had added a null check here before the for loop.

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+        return withTransport(clazz, configBuilder.build());
+    }
+
+    public <T extends ClientTransport> ClientBuilder withTransport(Class<T> clazz, ClientTransportConfig<T> config) {
+        clientTransports.put(clazz, config);
+
+        return this;
+    }
+
+    public ClientBuilder useClientPreference(boolean useClientPreference) {
```
fjuma: Currently, this still also appears in ClientConfig. Just wondering if it's needed in both places or if we can just have it in one of them?

README.md
brasseld: I did this because jsonrpc does not require 3rd dependencies to work. This is obviously not the case for gRPC which is requiring the gRPC api.

Also, from a developer experience perspective, it's easier to only rely on a single depency to get things work.

README.md
brasseld: Yep, it makes sense for that one !

client/transport/spi/src/main/java/io/a2a/client/transport/spi/ClientTransportConfigBuilder.java
brasseld: Good point, these are being added automatically, I will remove them

README.md
brasseld: My only concern with one is that it will force to have the Builder class as an inner class to the Client. 
Do you think this one is the huge issue ?

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransportConfig.java
brasseld: Lazy loaded from the builder: https://github.com/a2aproject/a2a-java/pull/238/files#diff-4fe99c4e576da4940f05f4cbbd8dffff5b47373dd17cf3abb76e93bfa2a12c19R20

I did this to avoid creating a uselss instance if one is providing its own implementation.

client/base/src/main/java/io/a2a/client/ClientBuilder.java
brasseld: Hum, not sure here. there is  only a single handler , so I don't think it makes sense to prefix it with `add`.

README.md
fjuma: That's the approach we've taken for other Builders but either approach is fine, i.e.,

`new Client.Builder()...` or `new ClientBuilder()...` instead of `Client.from`

client/base/src/main/java/io/a2a/client/ClientBuilder.java
brasseld: Actually, it is not entirely true.
You can only add one for each transport., that's the reason why I prefer I choose to go for `with `

client/base/src/main/java/io/a2a/client/ClientBuilder.java
fjuma: Ah, right, that's fine. Maybe we could just rename to `streamingErrorHandler`

client/base/src/main/java/io/a2a/client/ClientBuilder.java
fjuma: Ok so maybe `add` makes sense for the general `withTransport` method since multiple custom transports could be configured?

README.md
```diff
@@ -305,14 +298,18 @@ Consumer<Throwable> errorHandler = error -> {
...
+        .streamErrorHandler(errorHandler)
+        .build();
 ```
 
 #### Configuring Transport-Specific Settings
 
-Different transport protocols can be configured with specific settings using `ClientTransportConfig` implementations. The A2A Java SDK provides `JSONRPCTransportConfig` for the JSON-RPC transport and `GrpcTransportConfig` for the gRPC transport.
+Different transport protocols can be configured with specific settings using specific `ClientTransportConfig` implementations. The A2A Java SDK provides `JSONRPCTransportConfig` for the JSON-RPC transport and `GrpcTransportConfig` for the gRPC transport.
 
 ##### JSON-RPC Transport Configuration
```
fjuma, line 317: The line at the beginning of this section will need to be updated since it currently says:

"For the JSON-RPC transport, if you'd like to use the default JdkA2AHttpClient, no additional configuration is needed..."

But with the current approach, `withJsonRpcTransport` must be called.

fjuma: Looks like there are still some references to ClientFactory in the README that need to be updated.

README.md
fjuma: Ok so with this approach, the JSON-RPC transport provider will be discovered during service loader discovery but will only get used if appropriate config has been specified via `withJsonRpcTransport`. We should update the "Sample Usage" section as mentioned below to reflect this.

brasseld:
@fjuma thanks for your comments.

Most (all !) the points you have raised should have been fixed now.

README.md
```diff
@@ -305,32 +300,39 @@ Consumer<Throwable> errorHandler = error -> {
...
 
-// Create the client using ClientFactory
-ClientFactory clientFactory = new ClientFactory(clientConfig);
-Client client = clientFactory.create(agentCard, consumers, errorHandler);
+// Create the client using the builder
+Client client = ClientBuilder
+        .agentCard(agentCard)
+        .clientConfig(clientConfig)
+        .addStreamConsumers(consumers)
+        .streamingErrorHandler(errorHandler)
```
fjuma, line 309: In this code snippet, I think we should add a transport, e.g.,

`.withTransport(JSONRPCTransport.class, new JSONRPCTransportConfigBuilder().build())`

README.md
```diff
@@ -305,32 +300,39 @@ Consumer<Throwable> errorHandler = error -> {
...
-JSONRPCTransportConfig jsonrpcConfig = new JSONRPCTransportConfig(customHttpClient);
-
-// Configure the client with transport-specific settings
+// Configure the client settings
 ClientConfig clientConfig = new ClientConfig.Builder()
         .setAcceptedOutputModes(List.of("text"))
-        .setClientTransportConfigs(List.of(jsonrpcConfig))
+        .build();
+
+Client client = ClientBuilder
```
fjuma: After thinking about this one some more, maybe we should add a static `builder()` method to `Client` that returns a `new ClientBuilder()` and then this could be updated to:

```
Client.builder()
   .agentCard(...)
   ...
```

WDYT?

client/base/pom.xml
```diff
@@ -20,15 +20,22 @@
...
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>a2a-java-sdk-http-client</artifactId>
+            <artifactId>a2a-java-sdk-client-transport-spi</artifactId>
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>a2a-java-sdk-client-transport-spi</artifactId>
+            <artifactId>a2a-java-sdk-client-transport-jsonrpc</artifactId>
+            <version>${project.version}</version>
```
fjuma: Just minor, the version can be removed here.

client/base/pom.xml
```diff
@@ -20,15 +20,22 @@
...
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>a2a-java-sdk-client-transport-grpc</artifactId>
+            <version>${project.version}</version>
```
fjuma: Just minor, the version can be removed here.

README.md
```diff
@@ -363,15 +368,11 @@ will be used based on the selected transport:
...
-    new GrpcTransportConfig(...)
-);
-
-ClientConfig clientConfig = new ClientConfig.Builder()
-        .setAcceptedOutputModes(List.of("text"))
-        .setClientTransportConfigs(transportConfigs)
-        .build();
+Client client = ClientBuilder
+                .agentCard(agentCard)
+                .withTransport(GrpcTransport.class, new GrpcTransportConfigBuilder().build())
```
fjuma: We should update to this to `new GrpcTransportConfigBuilder().channelFactory(...).build()` since a channel factory must be specified.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfig.java
```diff
@@ -1,11 +1,11 @@
...
-import java.util.function.Function;
-
-import io.a2a.client.config.ClientTransportConfig;
+import io.a2a.client.transport.spi.ClientTransportConfig;
 import io.grpc.Channel;
 
-public class GrpcTransportConfig implements ClientTransportConfig {
+import java.util.function.Function;
+
+public class GrpcTransportConfig extends ClientTransportConfig<GrpcTransport> {
```
fjuma, line 9: The constructor should have a `checkNotNullParam` call since the channel factory must be specified.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
```diff
@@ -0,0 +1,22 @@
...
+import io.grpc.Channel;
+
+import java.util.function.Function;
+
+public class GrpcTransportConfigBuilder extends ClientTransportConfigBuilder<GrpcTransportConfig, GrpcTransportConfigBuilder> {
+
+    private Function<String, Channel> channelFactory;
+
+    public GrpcTransportConfigBuilder channelFactory(Function<String, Channel> channelFactory) {
+        this.channelFactory = channelFactory;
```
fjuma, line 16: Can add a `checkNotNullParam` call here.

client/base/src/test/java/io/a2a/client/ClientBuilderTest.java
```diff
@@ -0,0 +1,84 @@
...
+            .build();
+
+    @Test()
+    public void shouldNotFindCompatibleTransport() throws A2AClientException {
+        A2AClientException exception = Assertions.assertThrows(A2AClientException.class,
+                () -> ClientBuilder
+                        .agentCard(card)
+                        .clientConfig(new ClientConfig.Builder().setUseClientPreference(true).build())
+                        .withTransport(GrpcTransport.class, new GrpcTransportConfigBuilder()
+                                .channelFactory(null))
```
fjuma: It shouldn't be possible to specify a null value for the channel factory.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportProvider.java
```diff
@@ -1,35 +1,29 @@
...
-                if (clientTransportConfig instanceof GrpcTransportConfig grpcTransportConfig) {
-                    Channel channel = grpcTransportConfig.getChannelFactory().apply(agentUrl);
-                    return new GrpcTransport(channel, agentCard);
-                }
-            }
+
+        Function<String, Channel> channelFactory = grpcTransportConfig.getChannelFactory();
+
+        if (channelFactory != null) {
+            Channel channel = grpcTransportConfig.getChannelFactory().apply(agentUrl);
```
fjuma: Since a GrpcTransportConfig must be specified to use gRPC, I think it makes sense to require the channel factory to be specified when creating that. So here, the channel factory should always be non-null.

http-client/pom.xml
```diff
@@ -17,6 +17,12 @@
     <description>Java SDK for the Agent2Agent Protocol (A2A) - HTTP Client</description>
 
     <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>a2a-java-sdk-spec</artifactId>
+            <version>${project.version}</version>
```
fjuma: Just minor, no need for the version here.

fjuma (changes requested):

README.md
brasseld: But we need to enforce that the dev is passing an agentCard, so that should probably be:

```java
Client.builder(agentCard) 
```

instead ?

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
```diff
@@ -0,0 +1,22 @@
+package io.a2a.client.transport.grpc;
+
+import io.a2a.client.transport.spi.ClientTransportConfigBuilder;
+import io.grpc.Channel;
+
+import java.util.function.Function;
+
+public class GrpcTransportConfigBuilder extends ClientTransportConfigBuilder<GrpcTransportConfig, GrpcTransportConfigBuilder> {
```
fjuma, line 9: Was just discussing these changes with some others and one thing I'm wondering is, do we really need builders for GrpcTransportConfig and JSONRPCTransportConfig since they each only have one thing that can be configured?

brasseld: That is a fair comment.

The current implementation is not strict with the use of a TransportBuilder, you can pass directly an instance of a TransportConfig when calling the `withTransport()` method.

brasseld: The ConfigBuilder is also supporting the addition of interceptors. 
See https://github.com/a2aproject/a2a-java/pull/238/files#diff-5313eab4c9eb6dcc4391572e14bfc8c76eb502e91ab48fdd329eae3faac336cfR13

brasseld:
I have pushed a new version to fix your previous comments, and also provided an implementation for this https://github.com/a2aproject/a2a-java/pull/238#discussion_r2301316208

README.md
fjuma: Looks good.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
fjuma: Ok makes sense. Maybe we can just simplify the README for now and update things like `.withTransport(JSONRPCTransport.class, new JSONRPCTransportConfigBuilder().build())` to just
`.withTransport(JSONRPCTransport.class, new JSONRPCTransportConfig())` and then once we add interceptor implementations we can add examples with the builder to show how to add them. Interceptors aren't supported for gRPC right now so there `new GrpcTransportConfig(...)` could be used.

That takes us back to this comment though, I think it would be good to be able to specify just `new JSONRPCTransportConfig()` and have the default HTTP client implementation get used.
https://github.com/wildfly-extras/a2a-java-sdk-server-jakarta/pull/16/files#diff-25d7ec0e7434a0c0b37388f5e3f918e4640d5a64a29189337682a721a6272851R38

brasseld: That makes sense @fjuma 

I've updated the PR / readme accordingly.

README.md
```diff
@@ -305,32 +300,39 @@ Consumer<Throwable> errorHandler = error -> {
...
 
-// Create the client using ClientFactory
-ClientFactory clientFactory = new ClientFactory(clientConfig);
-Client client = clientFactory.create(agentCard, consumers, errorHandler);
+// Create the client using the builder
+Client client = Client
+        .builder(agentCard)
+        .clientConfig(clientConfig)
+        .withTransport(JSONRPCTransport.class, new JSONRPCTransportConfig())
+        .addStreamConsumers(consumers)
```
fjuma: .addConsumers(...)

README.md
```diff
@@ -346,13 +348,15 @@ Function<String, Channel> channelFactory = agentUrl -> {
...
 // Configure the client with transport-specific settings
 ClientConfig clientConfig = new ClientConfig.Builder()
         .setAcceptedOutputModes(List.of("text"))
-        .setClientTransportConfigs(List.of(grpcConfig))
+        .build();
+
+Client client = Client
+        .builder(agentCard)
+        .clientConfig(clientConfig)
+        .withTransport(GrpcTransport.class, new GrpcTransportConfigBuilder(channelFactory))
```
fjuma: `new GrpcTransportConfig(channelFactory)`

fjuma:
Thanks very much for all the updates, @brasseld! I've just added a couple final comments for the README.

brasseld:
Last updates puhsed @fjuma

fjuma:
@kabir @ehsavoie Please review when you get a chance, thanks!

ehsavoie:
I understand there is a 'global' ClientConfig to be shared across transports, but I'm wondering what  would be the case if we want a different configuration per transport. It feels a bit strange to define a configuration for the client to then define one per transport.
Otherwise LGTM

kabir:
@ehsavoie @fjuma Just throwing this out there without much thought. But I think perhaps if what you say becomes an issue, then an option in the future could be:

* ClientConfig becomes the global/default setting
* The transport configs are enhanced to allow setting the same stuff as in ClientConfig, so those become overrides

Then when constructing the client for each transport, we merge in the overrides with the global ClientConfig for each transport

brasseld:
@kabir @fjuma @ehsavoie 
I would say that, generally speaking, the transport configuration is something really specific to the transport itself while everything in the ClientConfig is more general but does only apply to the client iself (ClientConfig is no more passed to the Transport, and there is now a clear separation of concerns).

I quickly looked at how it's done for the MCP SDK, and it's somehow similar approach: https://modelcontextprotocol.io/sdk/java/mcp-client#sse-httpclient

ehsavoie:
@brasseld It was more a theorical question, and we can rethink about it if this question is raised in the future.

client/base/src/main/java/io/a2a/client/config/ClientConfig.java
```diff
@@ -12,21 +12,18 @@ public class ClientConfig {
...
 
-    public ClientConfig(Boolean streaming, Boolean polling, List<ClientTransportConfig> clientTransportConfigs,
-                        List<String> supportedTransports, Boolean useClientPreference,
+    public ClientConfig(Boolean streaming, Boolean polling, List<String> supportedTransports, Boolean useClientPreference,
                         List<String> acceptedOutputModes, PushNotificationConfig pushNotificationConfig,
                         Integer historyLength, Map<String, Object> metadata) {
         this.streaming = streaming == null ? true : streaming;
         this.polling = polling == null ? false : polling;
-        this.clientTransportConfigs = clientTransportConfigs;
         this.supportedTransports = supportedTransports;
```
fjuma: Was just looking closer at `ClientConfig` after Emmanuel's question and just realized that we should remove `supportedTransports` from here since that's now determined by the transports that are configured with `withTransport`. WDYT?

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -0,0 +1,160 @@
...
+        }
+        return serverPreferredTransports;
+    }
+
+    private List<String> getClientPreferredTransports() {
+        List<String> supportedClientTransports = new ArrayList<>();
+
+        if (clientTransports.isEmpty()) {
+            // default to JSONRPC if not specified
+            supportedClientTransports.add(TransportProtocol.JSONRPC.asString());
```
fjuma, line 122: Same here, looking into `ClientConfig` made me realize that if we get here, no transports have been configured with `withTransport` so the call to retrieve the configuration associated with the transport later on will return null.

brasseld: Good catch @fjuma , I've updated the PR and added a unit test to cover this.

fjuma:
Thanks @brasseld!

fjuma:
@kabir @ehsavoie Thanks for the comments! I agree that we can always revisit the configuration if needed in the future. I'll go ahead and merge this.

PR #241, opened 2025-08-27
fix: duplication of parameters in json for defining subtypes.

### LINKED ISSUE #234: [Bug]: a2a request body: multiple duplicate "kind" values appear after serialization
### What happened?

When Jackson serializes a class that implements multiple interfaces:
The getKind() method defined for each interface will be regarded as an independent attribute source by Jackson.
Both the `EventKind `and `StreamingEventKind `interfaces define the getKind() method. This is the fundamental reason for the occurrence of duplicate kind attributes during serialization.

### Relevant log output

```shell
{"kind":"message","role":"user","parts":[{"kind":"text","text":"please help me plan the route from Tian'anmen Square to Beijing West Railway Station","kind":"text"}],"messageId":"68a29ed35be55049f4350d07","contextId":"1755485011723f65lx3qvj","metadata":{"userId":"4b1fecf5-ee6a-4e1f-a122-75f5206e96fb"},"kind":"message"}
In the above report, the word "kind":"text"/"message" was repeated twice.
```

### Code of Conduct

xumk:
fixed #234 fix kind dublicate because classes have gatKind method and add correctly JsonTypeName for type

### DIFF
--- a/spec/src/main/java/io/a2a/spec/DataPart.java
+++ b/spec/src/main/java/io/a2a/spec/DataPart.java
@@ -6,15 +6,20 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.a2a.util.Assert;
+
+import static io.a2a.spec.DataPart.DATA;
 
 /**
  * Represents a structured data segment (e.g., JSON) within a message or artifact.
  */
+@JsonTypeName(DATA)
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class DataPart extends Part<Map<String, Object>> {
 
+    public static final String DATA = "data";
     private final Map<String, Object> data;
     private final Map<String, Object> metadata;
     private final Kind kind;
--- a/spec/src/main/java/io/a2a/spec/FilePart.java
+++ b/spec/src/main/java/io/a2a/spec/FilePart.java
@@ -6,16 +6,21 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.a2a.util.Assert;
+
+import static io.a2a.spec.FilePart.FILE;
 
 /**
  * Represents a file segment within a message or artifact. The file content can be
  * provided either directly as bytes or as a URI.
  */
+@JsonTypeName(FILE)
 @JsonInclude(JsonInclude.Include.NON_ABSENT)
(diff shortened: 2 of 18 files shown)


fjuma:
@xumk Thanks for the PR! It looks like there are some test failures related to the change.

xumk:
@fjuma yes. i fixed it

fjuma:
@xumk It would be good to add tests for this change.

xumk:
@fjuma I'll try to do it soon.

xumk:
@fjuma Hi, I added the tests.

spec/src/main/java/io/a2a/spec/DataPart.java
```diff
@@ -6,11 +6,13 @@
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.a2a.util.Assert;
 
 /**
  * Represents a structured data segment (e.g., JSON) within a message or artifact.
  */
+@JsonTypeName("data")
```
fjuma: Would be good to use a constant here and then update the other places that use "data" to also use the constant.

spec/src/main/java/io/a2a/spec/FilePart.java
```diff
@@ -6,12 +6,14 @@
...
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.a2a.util.Assert;
 
 /**
  * Represents a file segment within a message or artifact. The file content can be
  * provided either directly as bytes or as a URI.
  */
+@JsonTypeName("file")
```
fjuma: We should use a constant here.

spec/src/main/java/io/a2a/spec/TextPart.java
```diff
@@ -6,11 +6,13 @@
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
 import io.a2a.util.Assert;
 
 /**
  * Represents a text segment within a message or artifact.
  */
+@JsonTypeName("text")
```
fjuma: Should use a constant here.

spec/src/test/java/io/a2a/spec/SingleKeyHashMap.java
```diff
@@ -0,0 +1,14 @@
+package io.a2a.spec;
+
+import java.util.HashMap;
+
+public class SingleKeyHashMap <K, V> extends HashMap<K, V> {
```
fjuma: Very minor, this could probably be a static class in the test class.

xumk:
@fjuma i added it

spec/src/main/java/io/a2a/spec/Part.java
```diff
@@ -12,7 +12,7 @@
  */
 @JsonTypeInfo(
```
fjuma, line 13: I think this class can now be updated to make use of the newly introduced constants.

xumk: sorry. I overlooked it. I Added it

fjuma: Thanks!

PR #251, opened 2025-09-03
feat: Check transports in AgentCard match those on classpath

### LINKED ISSUE #246: Log a warning if the transports specified in the AgentCard don't match the transports that are available on the classpath
The transports that are available for an A2A server agent are determined by the dependencies that have been added to the application, e.g., see https://github.com/a2aproject/a2a-java/blob/main/README.md#server-transports.

When creating an `AgentCard`, the user must specify the primary endpoint for their A2A server agent and if the server agent supports multiple transports, these need to be specified as additional interfaces in the `AgentCard`.

It might be good to log some kind of warning if the `AgentCard` does not include all the transports available on the classpath.

Similarly, if an additional interface has been specified for a transport that isn't available on the classpath, it might be good to also log a warning in this case.

kabir:
Fixes #246  🦕

### DIFF
--- a/server-common/src/main/java/io/a2a/server/TransportMetadata.java
+++ b/server-common/src/main/java/io/a2a/server/TransportMetadata.java
@@ -13,7 +13,7 @@ public interface TransportMetadata {
      * 
      * @return the transport protocol
      */
-    TransportProtocol getTransportProtocol();
+    String getTransportProtocol();
 
     /**
      * Checks if this transport endpoint is currently available/functional.
--- a/reference/grpc/src/main/java/io/a2a/server/grpc/quarkus/QuarkusGrpcTransportMetadata.java
+++ b/reference/grpc/src/main/java/io/a2a/server/grpc/quarkus/QuarkusGrpcTransportMetadata.java
@@ -5,7 +5,7 @@ import io.a2a.spec.TransportProtocol;
 
 public class QuarkusGrpcTransportMetadata implements TransportMetadata {
     @Override
-    public TransportProtocol getTransportProtocol() {
-        return TransportProtocol.GRPC;
+    public String getTransportProtocol() {
+        return TransportProtocol.GRPC.asString();
     }
 }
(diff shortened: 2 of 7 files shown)


tests/server-common/src/test/java/io/a2a/server/apps/common/AgentCardProducer.java
```diff
@@ -33,8 +42,33 @@ public AgentCard agentCard() {
...
+                properties.load(in);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        String preferredTransport = properties.getProperty(PREFERRED_TRANSPORT);
+        Assertions.assertNotNull(preferredTransport);
+        return preferredTransport;
     }
```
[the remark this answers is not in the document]
kabir: I don't think this makes sense. It is only test code, and dependencies between those is a bit tricky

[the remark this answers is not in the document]
kabir: Won't fix

server-common/src/main/java/io/a2a/server/TransportMetadata.java
```diff
@@ -0,0 +1,27 @@
...
+ * This is used by the validation system to discover available transports on the classpath.
+ */
+public interface TransportMetadata {
+    
+    /**
+     * Returns the transport protocol this endpoint supports.
+     * 
+     * @return the transport protocol
+     */
+    TransportProtocol getTransportProtocol();
```
fjuma: `TransportProtocol` is an `enum` that specifies the transports defined in the A2A spec but the spec also allows custom transports to be specified. I think we should use a String here instead of the enum. That's the approach we went with for `getTransportProtocol` in the `ClientTransportProvider` interface.

reference/jsonrpc/src/main/java/io/a2a/server/apps/quarkus/A2AServerRoutes.java
```diff
@@ -52,6 +54,7 @@
 import io.a2a.spec.SetTaskPushNotificationConfigRequest;
 import io.a2a.spec.StreamingJSONRPCRequest;
 import io.a2a.spec.TaskResubscriptionRequest;
+import io.a2a.spec.TransportProtocol;
```
fjuma: Are these imports needed here? It doesn't look like they're used. Or is it intentional as I also see these in QuarkusGrpcHandler?

server-common/src/main/java/io/a2a/server/TransportMetadata.java
kabir: Good catch!

kabir:
@fjuma Should be fixed now

PR #252, opened 2025-09-03
feat: Use main HTTP port for gRPC

### LINKED ISSUE #250: Update the gRPC tests and TCK configuration to use Quarkus "vertx-based" gRPC
We need to set the following property in application.properties and then update test configuration accordingly to use the HTTP port instead of a separate port for gRPC:

```
# Use the Vert.x based server
quarkus.grpc.server.use-separate-server=false
```

kabir:
Fixes #250 🦕

### DIFF
--- a/reference/grpc/src/test/java/io/a2a/server/grpc/quarkus/QuarkusA2AGrpcTest.java
+++ b/reference/grpc/src/test/java/io/a2a/server/grpc/quarkus/QuarkusA2AGrpcTest.java
@@ -29,7 +29,9 @@ public class QuarkusA2AGrpcTest extends AbstractA2AServerTest {
 
     @Override
     protected String getTransportUrl() {
-        return "localhost:8081"; // gRPC server runs on port 9001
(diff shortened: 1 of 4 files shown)


reference/grpc/src/test/java/io/a2a/server/grpc/quarkus/QuarkusA2AGrpcTest.java
```diff
@@ -29,7 +29,7 @@ protected String getTransportProtocol() {
 
     @Override
     protected String getTransportUrl() {
-        return "localhost:9001"; // gRPC server runs on port 9001
+        return "localhost:8081"; // gRPC server runs on port 9001
```
fjuma: Just minor, we should fix this comment.

fjuma:
@kabir There are TCK failures here.

kabir:
@fjuma fixed. The TCK failures seem to be due to a flaky test. They pass again after just fixing the comment

PR #263, opened 2025-09-11
fix: Use deployment CL to load TransportMetadata services if different from TCCL

kabir:
I found working on https://github.com/wildfly-extras/a2a-java-sdk-server-jakarta/pull/26 that in WildFly gRPC happens on a separate thread with its own classloader. Hence the serviceloader lookup from the default TCCL didn't work in WildFLy.

### DIFF
--- a/transport/grpc/src/main/java/io/a2a/transport/grpc/handler/GrpcHandler.java
+++ b/transport/grpc/src/main/java/io/a2a/transport/grpc/handler/GrpcHandler.java
@@ -380,10 +380,33 @@ public abstract class GrpcHandler extends A2AServiceGrpc.A2AServiceImplBase {
     private AgentCard getAgentCardInternal() {
         AgentCard agentCard = getAgentCard();
         if (initialised.compareAndSet(false, true)) {
-            // Validate transport configuration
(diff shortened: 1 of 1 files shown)


transport/grpc/src/main/java/io/a2a/transport/grpc/handler/GrpcHandler.java
```diff
@@ -380,12 +380,34 @@ private <V> void handleInternalError(StreamObserver<V> responseObserver, Throwab
...
+        try {
+            // Set TCCL to the classloader that loaded this class, which should have access
+            // to the deployment classpath containing META-INF/services files
+            ClassLoader deploymentClassLoader = this.getClass().getClassLoader();
+            Thread.currentThread().setContextClassLoader(deploymentClassLoader);
+            
+            AgentCardValidator.validateTransportConfiguration(agentCard);
+        } catch (Exception e) {
+            // Log the error but don't fail if validation fails due to other issues
+            System.err.println("Warning: AgentCard transport validation failed: " + e.getMessage());
```
ehsavoie: Why not using a logger here ?

PR #285, opened 2025-09-22
feat: ServerCallContext ports from Python implementation

kabir:
Port server-side extension support and implement complete gRPC context access equivalent to Python's ServicerContext. Adds client/server metadata support, rich context information, and interceptor infrastructure while maintaining backward compatibility.

### DIFF
--- a/client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
+++ b/client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
@@ -61,11 +61,9 @@ public class GrpcTransport implements ClientTransport {
         checkNotNullParam("request", request);
 
         SendMessageRequest sendMessageRequest = createGrpcSendMessageRequest(request, context);
-        Metadata metadata = createGrpcMetadata(context);
 
         try {
-            // Create a stub with metadata attached
-            A2AServiceBlockingV2Stub stubWithMetadata = blockingStub.withInterceptors(MetadataUtils.newAttachHeadersInterceptor(metadata));
+            A2AServiceBlockingV2Stub stubWithMetadata = createBlockingStubWithMetadata(context);
             SendMessageResponse response = stubWithMetadata.sendMessage(sendMessageRequest);
             if (response.hasMsg()) {
                 return FromProto.message(response.getMsg());
@@ -85,12 +83,10 @@ public class GrpcTransport implements ClientTransport {
         checkNotNullParam("request", request);
         checkNotNullParam("eventConsumer", eventConsumer);
         SendMessageRequest grpcRequest = createGrpcSendMessageRequest(request, context);
-        Metadata metadata = createGrpcMetadata(context);
         StreamObserver<StreamResponse> streamObserver = new EventStreamObserver(eventConsumer, errorConsumer);
 
         try {
-            // Create a stub with metadata attached
-            A2AServiceStub stubWithMetadata = asyncStub.withInterceptors(MetadataUtils.newAttachHeadersInterceptor(metadata));
+            A2AServiceStub stubWithMetadata = createAsyncStubWithMetadata(context);
             stubWithMetadata.sendStreamingMessage(grpcRequest, streamObserver);
         } catch (StatusRuntimeException e) {
             throw GrpcErrorMapper.mapGrpcError(e, "Failed to send streaming message request: ");
@@ -109,7 +105,8 @@ public class GrpcTransport implements ClientTransport {
         GetTaskRequest getTaskRequest = requestBuilder.build();
 
         try {
-            return FromProto.task(blockingStub.getTask(getTaskRequest));
+            A2AServiceBlockingV2Stub stubWithMetadata = createBlockingStubWithMetadata(context);
+            return FromProto.task(stubWithMetadata.getTask(getTaskRequest));
         } catch (StatusRuntimeException e) {
             throw GrpcErrorMapper.mapGrpcError(e, "Failed to get task: ");
         }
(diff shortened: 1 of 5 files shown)


client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
```diff
@@ -59,9 +61,12 @@ public EventKind sendMessage(MessageSendParams request, ClientCallContext contex
         checkNotNullParam("request", request);
 
         SendMessageRequest sendMessageRequest = createGrpcSendMessageRequest(request, context);
+        Metadata metadata = createGrpcMetadata(context);
 
         try {
-            SendMessageResponse response = blockingStub.sendMessage(sendMessageRequest);
+            // Create a stub with metadata attached
+            A2AServiceBlockingV2Stub stubWithMetadata = blockingStub.withInterceptors(MetadataUtils.newAttachHeadersInterceptor(metadata));
```
fjuma: Just to check, the metadata is only being used for sendMessage and sendStreamingMessage. Is that intentional? Should it be used for all client requests?

server-common/src/main/java/io/a2a/server/extensions/A2AExtensions.java
```diff
@@ -0,0 +1,52 @@
...
+import io.a2a.common.A2AHeaders;
+import io.a2a.spec.AgentCard;
+import io.a2a.spec.AgentExtension;
+
+public class A2AExtensions {
+    /**
+     * HTTP header name for A2A extensions.
+     * @deprecated Use {@link A2AHeaders#X_A2A_EXTENSIONS} instead
+     */
+    @Deprecated
```
fjuma: This annotation can be removed right?

server-common/src/main/java/io/a2a/server/extensions/A2AExtensions.java
```diff
@@ -0,0 +1,52 @@
...
+    public static final String HTTP_EXTENSION_HEADER = A2AHeaders.X_A2A_EXTENSIONS;
```
fjuma: Is this needed?

transport/grpc/src/main/java/io/a2a/transport/grpc/handler/GrpcHandler.java
```diff
@@ -312,17 +318,53 @@ private <V> ServerCallContext createCallContext(StreamObserver<V> responseObserv
...
+            
+            // Extract requested extensions from gRPC context (set by interceptor)
+            Set<String> requestedExtensions = new HashSet<>();
+            try {
+                // Try to get extensions from gRPC context (available when interceptor is used)
+                String extensionsHeader = getExtensionsFromContext();
+                if (extensionsHeader != null) {
+                    requestedExtensions = A2AExtensions.getRequestedExtensions(List.of(extensionsHeader));
+                }
+            } catch (Exception e) {
```
fjuma: Can an exception actually be caught here? It doesn't look like getExtensionsFromContext() or getRequestedExtensions() throw an exception. Let me know if I'm missing something.

transport/grpc/src/main/java/io/a2a/transport/grpc/handler/GrpcHandler.java
```diff
@@ -418,4 +460,67 @@ public static void setStreamingSubscribedRunnable(Runnable runnable) {
...
+     * @return the peer information, or null if not available
+     */
+    protected static String getCurrentPeerInfo() {
+        try {
+            return GrpcContextKeys.PEER_INFO_KEY.get();
+        } catch (Exception e) {
+            return null;
+        }
+    }
 }
```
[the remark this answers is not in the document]
kabir: Fixed

kabir:
Thanks @fjuma. I've rebased and addressed the comments

reference/jsonrpc/src/main/java/io/a2a/server/apps/quarkus/A2AServerRoutes.java
```diff
@@ -241,7 +244,11 @@ public String getUsername() {
             headerNames.forEach(name -> headers.put(name, rc.request().getHeader(name)));
             state.put("headers", headers);
 
-            return new ServerCallContext(user, state);
+        // Extract requested extensions from X-A2A-Extensions header
+        List<String> extensionHeaderValues = rc.request().headers().getAll(A2AHeaders.X_A2A_EXTENSIONS);
```
fjuma: Very minor, indentation seems off here.

reference/rest/src/main/java/io/a2a/server/rest/quarkus/A2AServerRoutes.java
```diff
@@ -308,7 +313,11 @@ public String getUsername() {
             headerNames.forEach(name -> headers.put(name, rc.request().getHeader(name)));
             state.put("headers", headers);
 
-            return new ServerCallContext(user, state);
+        // Extract requested extensions from X-A2A-Extensions header
+        List<String> extensionHeaderValues = rc.request().headers().getAll(A2AHeaders.X_A2A_EXTENSIONS);
```
fjuma: Very minor, indentation seems off here.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
```diff
@@ -234,6 +245,50 @@ private SendMessageRequest createGrpcSendMessageRequest(MessageSendParams messag
...
+        
+        if (context != null && context.getHeaders() != null) {
+            // Set X-A2A-Extensions header if present
+            String extensionsHeader = context.getHeaders().get(A2AHeaders.X_A2A_EXTENSIONS);
+            if (extensionsHeader != null) {
+                Metadata.Key<String> extensionsKey = Metadata.Key.of(A2AHeaders.X_A2A_EXTENSIONS, Metadata.ASCII_STRING_MARSHALLER);
+                metadata.put(extensionsKey, extensionsHeader);
+            }
+            
+            // Add other headers as needed in the future
```
fjuma, line 263: Was just thinking some more about this and realized that I can probably modify https://github.com/a2aproject/a2a-java/pull/292 to add the authorization/api-key headers for gRPC in a similar way.

Currently, we are relying on the user configuring a gRPC `ClientInterceptor` to specify the token/API key. Instead, I think I can update `GrpcTransport` to make use of the `ClientCallInterceptors` so that the `AuthInterceptor` could be used for gRPC like we use it for the other transports to add the appropriate header. And then, I can update this method to add the appropriate header to the gRPC `metadata` if present.

PR #309, opened 2025-09-29
feat: Replicated Queue Manager based on Mp Reactive Messaging with Kafka

kabir:
It is implemented in such a way that the replication mechanism can be swapped with something else


extras/queue-manager-replicated/core/src/main/java/io/a2a/extras/queuemanager/replicated/core/ReplicatedQueueManager.java
```diff
@@ -0,0 +1,122 @@
...
+    private final ThreadLocal<Boolean> isHandlingReplicatedEvent = new ThreadLocal<>();
+
+    @Inject
+    private ReplicationStrategy replicationStrategy;
+
+    public ReplicatedQueueManager() {
+        this.delegate = new InMemoryQueueManager(new ReplicatingEventQueueFactory());
+    }
+
+    // For testing
```
fjuma, line 31: Just wondering, if this is for testing, does it need to be public?

PR #331, opened 2025-10-09
fix: Support multiple messages to same task after non-blocking sendMe…

kabir:
…ssage

Add test for non-blocking multi-message pattern where multiple sendMessage calls to same taskId continue processing. Test uncovered EventConsumer polling loops never terminating, causing resource leaks. Fix: Close EventQueue at appropriate times to stop infinite polling:
- EventConsumer.close() explicitly closes queue
- InMemoryQueueManager.close() fails fast on missing queue
- DefaultRequestHandler cleanup: streaming (always close) vs non-blocking (close only if task final, for multi-message support)
- Refactor ResultAggregator to use AsyncUtils.consumer() consistently

### DIFF
--- a/server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
+++ b/server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
@@ -479,7 +479,7 @@ public class DefaultRequestHandler implements RequestHandler {
                     } else {
                         // Only close queue if task is in a final state
                         Task task = taskStore.get(taskId);
-                        if (task != null && task.getStatus() != null && task.getStatus().state().isFinal()) {
+                        if (task != null && task.getStatus().state().isFinal()) {
                             queue.close();
                         } else {
                             LOGGER.debug("Task {} not in final state or not yet created, keeping queue open", taskId);
@@ -531,7 +531,7 @@ public class DefaultRequestHandler implements RequestHandler {
                         // For non-final states, queue must stay open for potential future messages to same taskId
                         // so we can handle the "fire and forget' case used e.g. in the TCK
                         Task task = taskStore.get(taskId);
-                        if (task != null && task.getStatus() != null && task.getStatus().state().isFinal()) {
+                        if (task != null && task.getStatus().state().isFinal()) {
                             LOGGER.debug("Task in final state, closing queue for task {}", taskId);
                             closeQueue = true;
                         } else {
@@ -590,6 +590,11 @@ public class DefaultRequestHandler implements RequestHandler {
                 pushSender.sendNotification(latestTask);
             }
         }
+    }
+
+    private boolean isTaskFinalInTaskStore(String taskId) {
+        Task task = taskStore.get(taskId);
+        return task != null && task.getStatus().state().isFinal();
     }
 
     private record MessageSendSetup(TaskManager taskManager, Task task, RequestContext requestContext) {}


server-common/src/main/java/io/a2a/server/tasks/ResultAggregator.java
```diff
@@ -114,21 +146,75 @@ public EventTypeAndInterrupt consumeAndBreakOnInterrupt(EventConsumer consumer,
...
+                        continueInBackground = true;
                     }
                     // For non-blocking calls, interrupt as soon as a task is available.
                     else if (!blocking) {
                         shouldInterrupt = true;
+                        continueInBackground = true;
+                    }
+                    // For blocking calls, interrupt when we get any task-related event (EventKind except Message)
+                    // Cancel subscription to free resources (client can resubscribe if needed)
+                    else if (blocking && (event instanceof EventKind && !(event instanceof Message))) {
```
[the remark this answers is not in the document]
kabir: Fixed, good catch

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -499,11 +508,45 @@ private void trackBackgroundTask(CompletableFuture<Void> task) {
...
+                            LOGGER.debug("Task not in final state, keeping queue open for task {}", taskId);
+                        }
+                    }
+
+                    if (closeQueue) {
+                        try {
+                            queueManager.close(taskId);
+                        } catch (Exception e) {
+                            LOGGER.debug("Error closing queue for task {}: {}", taskId, e.getMessage());
+                        }
```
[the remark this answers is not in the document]
kabir: Fixed

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -499,11 +509,48 @@ private void trackBackgroundTask(CompletableFuture<Void> task) {
...
+                        // For non-streaming calls, only close queue if task is in final state
+                        // For non-final states, queue must stay open for potential future messages to same taskId
+                        // so we can handle the "fire and forget' case used e.g. in the TCK
+                        Task task = taskStore.get(taskId);
+                        if (task != null && task.getStatus() != null && task.getStatus().state().isFinal()) {
+                            LOGGER.debug("Task in final state, closing queue for task {}", taskId);
+                            closeQueue = true;
+                        } else {
+                            LOGGER.debug("Task not in final state, keeping queue open for task {}", taskId);
+                        }
```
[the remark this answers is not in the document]
kabir: I think it is clearer the way I have it. I removed the `task.getStatus() != null` bit though since it isn't necessary

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -545,5 +592,10 @@ private void sendPushNotification(String taskId, ResultAggregator resultAggregat
         }
     }
 
+    private boolean isTaskFinalInTaskStore(String taskId) {
```
ehsavoie, line 595: It is not used, should we remove it ?

server-common/src/main/java/io/a2a/server/events/InMemoryQueueManager.java
```diff
@@ -57,7 +64,14 @@ public EventQueue createOrTap(String taskId) {
             // Make sure an existing queue has not been added in the meantime
             existing = queues.putIfAbsent(taskId, newQueue);
         }
-        return existing == null ? newQueue : existing.tap();
+        EventQueue result = existing == null ? newQueue : existing.tap();
+        if (existing == null) {
```
fjuma, line 68: Just to check, can this be null here? Let me know if I'm missing something.

transport/jsonrpc/src/test/java/io/a2a/transport/jsonrpc/handler/JSONRPCHandlerTest.java
```diff
@@ -72,6 +72,7 @@
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
```
fjuma, line 75: Just minor, looks like this isn't needed right?

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
fjuma: Looks like this could be used on lines 481 and 533

kabir:
Replaced by https://github.com/a2aproject/a2a-java/pull/333

@amithsourya ^^ contains this and the second half of what I was mentioning. It undos some of the changes in this PR

misysin:
@kabir It took me a while to wrap my head around everything that's happening in #333 & #331, but I can really appreciate the elegance of the solution. This is exactly the kind of sophisticated code I hope to write someday 😃

PR #341, opened 2025-10-12
test: add ToProto and FromProto tests

### LINKED ISSUE #304: [chore]: Add more tests to ToProtoTest
The `ProtoUtils` class contains classes for converting gRPC classes and spec classes.

We have a test class called `ToProtoTest` that has a very limited set of tests for this.

This issue is to add more test cases to thoroughly test `ProtoUtils.ToProto` and `ProtoUtils.FromProto`.

The corresponding tests in the Python SDK can be seen here:
https://github.com/a2aproject/a2a-python/blob/main/tests/utils/test_proto_utils.py

We should add more test cases similar to what's in the Python tests.

Akanksha928:
# Description

This PR adds new unit tests to improve coverage for `ProtoUtils.ToProto` and `ProtoUtils.FromProto`.

### Tests Added
- `testFromProtoPartUnsupportedType`: verifies invalid proto parts throw `InvalidParamsError`.
- `testTaskQueryParamsInvalidName`: makes sure invalid task name formats are handled gracefully.
- `testToProtoPartUnsupportedType`: checks handling of unsupported `Part` types during conversion.

Fixes #304

### DIFF
--- a/spec-grpc/src/test/java/io/a2a/grpc/utils/FromProtoTest.java
+++ b/spec-grpc/src/test/java/io/a2a/grpc/utils/FromProtoTest.java
@@ -1,0 +1,37 @@
+package io.a2a.grpc.utils;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+public class FromProtoTest {
+    @Test
+    public void testFromProtoPartUnsupportedType() {
+        io.a2a.grpc.Part emptyPart = io.a2a.grpc.Part.newBuilder().build();
+        io.a2a.grpc.Message invalidMessage = io.a2a.grpc.Message.newBuilder()
+                                                                .addContent(emptyPart)
+                                                                .build();
+        io.a2a.spec.InvalidParamsError exception = assertThrows(
+                io.a2a.spec.InvalidParamsError.class,
+                () -> ProtoUtils.FromProto.message(invalidMessage)
+        );
+
+        assertEquals("Invalid parameters", exception.getMessage());
+
+    }
+
+    @Test
+    public void testTaskQueryParamsInvalidName() {
+        io.a2a.grpc.GetTaskRequest request = io.a2a.grpc.GetTaskRequest.newBuilder()
+                                                                       .setName("invalid-name-format")
+                                                                       .build();
+
+        var result = ProtoUtils.FromProto.taskQueryParams(request);
+
+        assertNotNull(result);
+        assertEquals("invalid-name-format", result.id());
+    }
+
+}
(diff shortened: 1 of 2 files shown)


Akanksha928:
Hi, I’ve added three initial tests that match the Python equivalent.

Could you please confirm if this direction looks good before I add a few more?

spec-grpc/src/test/java/io/a2a/grpc/utils/FromProtoTest.java
```diff
@@ -0,0 +1,37 @@
...
+        io.a2a.grpc.Part emptyPart = io.a2a.grpc.Part.newBuilder().build();
+        io.a2a.grpc.Message invalidMessage = io.a2a.grpc.Message.newBuilder()
+                                                                .addContent(emptyPart)
+                                                                .build();
+        io.a2a.spec.InvalidParamsError exception = assertThrows(
+                io.a2a.spec.InvalidParamsError.class,
+                () -> ProtoUtils.FromProto.message(invalidMessage)
+        );
+
+        assertEquals("Invalid parameters", exception.getMessage());
```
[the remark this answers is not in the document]
fjuma: You can ignore this Gemini message, I think you're correct that an InvalidParamsError will occur.

spec-grpc/src/test/java/io/a2a/grpc/utils/FromProtoTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+    }
+
+    @Test
+    public void testTaskQueryParamsInvalidName() {
+        io.a2a.grpc.GetTaskRequest request = io.a2a.grpc.GetTaskRequest.newBuilder()
+                                                                       .setName("invalid-name-format")
+                                                                       .build();
+
+        var result = ProtoUtils.FromProto.taskQueryParams(request);
```
fjuma, line 31: Looking at the python test, it looks like it's expecting an error to occur in this case. If we're not currently getting an error, that likely means we should update the taskQueryParams method to make sure the name is in the right format.

spec-grpc/src/test/java/io/a2a/grpc/utils/ToProtoTest.java
```diff
@@ -289,4 +289,28 @@ public void convertTaskTimestampStatus() {
...
+
+        io.a2a.spec.Message message = new io.a2a.spec.Message.Builder()
+                .messageId("msg-unsupported")
+                .contextId("ctx-unsupported")
+                .role(io.a2a.spec.Message.Role.USER)
+                .parts(List.of(fakePart))
+                .build();
+        io.a2a.grpc.Message result = ProtoUtils.ToProto.message(message);
+
+        assertNotNull(result);
```
fjuma, line 312: Looking at the Python test, they are expecting an error to occur in this case. If we're not hitting an error here, that means that we likely need to update ToProto so that we get an error in this case.

fjuma:
@Akanksha928 It looks like you're on the right track! I've added some comments. Feel free to let us know if you have more questions. Thanks!

ehsavoie:
I wrongly merged that so I reverted @Akanksha928 May you please fix the comments from @fjuma and re-send a PR. Sorry for this

PR #342, opened 2025-10-13
chore: add Jspecify NullCheck in grpc transport (Fixes #334)

### LINKED ISSUE #334: [chore]: Add Jspecify NullCheck in the grpc transport
Add a package-info.java in each package and annotate it with @NullMarked and ensure that null values are properly defined.

devcom33:
This PR adds Jspecify nullness annotations to the gRPC transport module:

- Added @NullMarked in package-info.java for the package.
- Annotated nullable parameters with @Nullable in GrpcTransport.java and related classes.
- Ensured all fields and method signatures match nullability expectations.
- Fixed NullAway errors by initializing non-null fields and updating constructor annotations.
- All tests pass successfully.

### DIFF
--- a/client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
+++ b/client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
@@ -6,11 +6,11 @@ import io.grpc.Channel;
 
 import java.util.function.Function;
 
+import org.jspecify.annotations.Nullable;
+
 public class GrpcTransportConfigBuilder extends ClientTransportConfigBuilder<GrpcTransportConfig, GrpcTransportConfigBuilder> {
 
-    private Function<String, Channel> channelFactory = s -> { 
-        throw new IllegalStateException("channelFactory must be set"); 
-    };
+    private @Nullable Function<String, Channel> channelFactory;
 
     public GrpcTransportConfigBuilder channelFactory(Function<String, Channel> channelFactory) {
         Assert.checkNotNullParam("channelFactory", channelFactory);
@@ -22,6 +22,9 @@ public class GrpcTransportConfigBuilder extends ClientTransportConfigBuilder<Grp
 
     @Override
     public GrpcTransportConfig build() {
+        if (channelFactory == null) {
+            throw new IllegalStateException("channelFactory must be set");
+        }
         GrpcTransportConfig config = new GrpcTransportConfig(channelFactory);
         config.setInterceptors(interceptors);
         return config;


ehsavoie:
@devcom33 may  you fix gemini comments please ?

heymouad:
Hey @ehsavoie , I’ve updated the builder to use the fail-fast pattern as suggested. 
Thanks for your feedback.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
```diff
@@ -60,26 +64,27 @@ public class GrpcTransport implements ClientTransport {
...
         this.interceptors = interceptors;
     }
 
     @Override
-    public EventKind sendMessage(MessageSendParams request, ClientCallContext context) throws A2AClientException {
+    public EventKind sendMessage(MessageSendParams request, @Nullable ClientCallContext context) throws A2AClientException {
         checkNotNullParam("request", request);
 
-        SendMessageRequest sendMessageRequest = createGrpcSendMessageRequest(request, context);
+        SendMessageRequest sendMessageRequest = createGrpcSendMessageRequest(request);
```
fjuma: `context` should be added back here.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
```diff
@@ -100,10 +105,10 @@ public EventKind sendMessage(MessageSendParams request, ClientCallContext contex
 
     @Override
     public void sendMessageStreaming(MessageSendParams request, Consumer<StreamingEventKind> eventConsumer,
-                                     Consumer<Throwable> errorConsumer, ClientCallContext context) throws A2AClientException {
+                                     Consumer<Throwable> errorConsumer, @Nullable ClientCallContext context) throws A2AClientException {
         checkNotNullParam("request", request);
         checkNotNullParam("eventConsumer", eventConsumer);
-        SendMessageRequest grpcRequest = createGrpcSendMessageRequest(request, context);
+        SendMessageRequest grpcRequest = createGrpcSendMessageRequest(request);
```
fjuma: Same here, `context` should be added back.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
```diff
@@ -270,7 +275,7 @@ public AgentCard getAgentCard(ClientCallContext context) throws A2AClientExcepti
     public void close() {
     }
 
-    private SendMessageRequest createGrpcSendMessageRequest(MessageSendParams messageSendParams, ClientCallContext context) {
+    private SendMessageRequest createGrpcSendMessageRequest(MessageSendParams messageSendParams) {
```
fjuma: We shouldn't remove context here.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransport.java
```diff
@@ -286,7 +287,7 @@ private SendMessageRequest createGrpcSendMessageRequest(MessageSendParams messag
      * Creates gRPC metadata from ClientCallContext headers.
      * Extracts headers like X-A2A-Extensions and sets them as gRPC metadata.
      */
-    private Metadata createGrpcMetadata(ClientCallContext context, PayloadAndHeaders payloadAndHeaders) {
+    private Metadata createGrpcMetadata(@Nullable ClientCallContext context, @Nullable PayloadAndHeaders payloadAndHeaders) {
```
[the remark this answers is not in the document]
fjuma: Would be good to update this.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
```diff
@@ -3,12 +3,13 @@
...
 import io.a2a.util.Assert;
 import io.grpc.Channel;
+import org.jspecify.annotations.Nullable;
 
 import java.util.function.Function;
 
 public class GrpcTransportConfigBuilder extends ClientTransportConfigBuilder<GrpcTransportConfig, GrpcTransportConfigBuilder> {
 
-    private Function<String, Channel> channelFactory;
+    private @Nullable Function<String, Channel> channelFactory;
```
fjuma: Since the `channelFactory` is required, we shouldn't make this nullable.

heymouad:
thank you @fjuma for your feedbacks
i've restored the context parameter in createGrpcSendMessageRequest method as you suggested.
Additionally, i removed the @Nullable annotation from channelFactory.

please let me know if there's anything else you need or if further changes are required.

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
```diff
@@ -8,7 +8,9 @@
 
 public class GrpcTransportConfigBuilder extends ClientTransportConfigBuilder<GrpcTransportConfig, GrpcTransportConfigBuilder> {
 
-    private Function<String, Channel> channelFactory;
+    private Function<String, Channel> channelFactory = s -> { 
+        throw new IllegalStateException("channelFactory must be set");
```
fjuma: I don't think this is needed.

heymouad: Without the initialization, it gives me NullAway fails at compile-time with:
`error: [NullAway] @NonNull field channelFactory not initialized`

That's why I initialized the lambda, it ensures the field is never null, satisfying NullAway's null-safety checks. Without this initialization, the build fails due to the @NullMarked annotation at the package level, which requires all fields to be non-null.

ehsavoie: I think we should consider the channelFactory field as Nullable since the constructor wouldn't set it, but we should add 
        Assert.checkNotNullParam("channelFactory", channelFactory); at the beginning of the build command.

heymouad: I tried `Assert.checkNotNullParam("channelFactory", channelFactory)` at the beginning of the `build()` method as you suggested, but NullAway still threw a compile error.
I think the issue is that NullAway doesn't recognize `Assert.checkNotNullParam` as a null-safety guarantee—it only understands explicit if (x == null) checks.

So that's why I tried an explicit null check:
`
@Override
public GrpcTransportConfig build() {
    if (channelFactory == null) {
        throw new IllegalStateException("channelFactory must be set");
    }
    GrpcTransportConfig config = new GrpcTransportConfig(channelFactory);
    config.setInterceptors(interceptors);
    return config;
}
` 
What do you think about this approach?

ehsavoie: Looks good to me, but i don't remember hitting the issue with Assert

heymouad: I understand that you didn't encounter the issue with Assert.checkNotNullParam. When I implemented it, but I received the following error from NullAway:

image

And here is the code I've tested:
public class GrpcTransportConfigBuilder extends ClientTransportConfigBuilder<GrpcTransportConfig, GrpcTransportConfigBuilder> {

    private @Nullable Function<String, Channel> channelFactory;

    public GrpcTransportConfigBuilder channelFactory(Function<String, Channel> channelFactory) {
        Assert.checkNotNullParam("channelFactory", channelFactory);

        this.channelFactory = channelFactory;

        return this;
    }

    @Override
    public GrpcTransportConfig build() {
        Assert.checkNotNullParam("channelFactory", channelFactory);
        GrpcTransportConfig config = new GrpcTransportConfig(channelFactory);
        config.setInterceptors(interceptors);
        return config;
    }
}

client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcTransportConfigBuilder.java
ehsavoie: The issue is that it is null when the builder instance is created. We need to ensure everything is ok at the build() call thus it is not null when passed to the constructor

fjuma:
Thanks @devcom33!

heymouad:
thank you for merging the PR.
would it be possible to add the hacktoberfest-accepted label to the PR? I'd really appreciate it!

fjuma:
@devcom33 Sure thing, I've added the label.

PR #348, opened 2025-10-14
fix: Implement background consumption for blocking, non-streaming onMessageSend requests

kabir:
Modify ResultAggregator.consumeAndBreakOnInterrupt to continue consuming
events in the background for blocking calls, matching the behavior of
non-blocking and streaming flows.

Previously, blocking calls set continueInBackground=false, which closed
the consumer after the first event was returned. This caused subsequent
events from the still-running AgentExecutor to be enqueued but never
processed, resulting in incomplete task state in the TaskStore.

The fix ensures:
- Blocking calls still return immediately with the first event
- Consumer continues processing events in background until agent completes
- All events are persisted to TaskStore, including final task state
- 
This matches the background processing pattern already established in
PR #284 for streaming client disconnects, but now applies to the
blocking non-streaming code path as well.

### DIFF
--- a/server-common/src/main/java/io/a2a/server/events/QueueManager.java
+++ b/server-common/src/main/java/io/a2a/server/events/QueueManager.java
@@ -1,6 +1,7 @@
 package io.a2a.server.events;
 
 public interface QueueManager {
+
     void add(String taskId, EventQueue queue);
 
     EventQueue get(String taskId);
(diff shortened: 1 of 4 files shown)


kabir:
This seems to introduce an intermittent failure in this test https://github.com/a2aproject/a2a-java/pull/348/files#diff-99b869ef9cc2557374d0ae2c161c8272f8dcbdbbad8fa7dd429cc2f03d9c6da2R71. I need to look into that, but want to finish the replicated queue manager work first

kabir:
Superseded by #351

Issue #373, opened 2025-10-20
Create a sample that shows how to create agent(s) deployed on k8s


fjuma:

kabir:
Done by #389

PR #387, opened 2025-10-22
feat: Add cloud deployment example with event replication

### LINKED ISSUE #281: Experiment with deploying an A2A server agent on k8s to identify any issues/gaps with being able to scale agents
It should be possible to have multiple instances of an A2A server agent. We need to investigate how load balancing/session affinity work in this case.

We should check to see if SmallRye Stork could be helpful for this.

kabir:
Add Kubernetes deployment example demonstrating multi-instance A2A agents with database persistence and Kafka-based event replication.

Fixes #281 and #373


kabir:
Replaced by #389

PR #431, opened 2025-11-03
fix: Wait for agent completion and ensure all events processed in blo…

### LINKED ISSUE #428: [Bug]: DefaultRequestHandler does not block until final state when blocking=true
## Description
version: v0.3.0.Final, v0.3.1.Final

When calling DefaultRequestHandler.onMessageSend with blocking=true, it is expected that the handler waits for the final state result before returning the response.
However, according to the test case below, the first message is returned immediately, and the remaining messages are processed in the background:

🔗 [DefaultRequestHandlerTest.java#L504](https://github.com/a2aproject/a2a-java/blob/5031b8faba7b78bbd508a3d21a6b1940ecf19013/server-common/src/test/java/io/a2a/server/requesthandlers/DefaultRequestHandlerTest.java#L504)

This behavior seems indistinguishable from the blocking=false case.

## Background

I understand that ResultAggregator was modified to run subscription logic on a background thread to prevent blocking the Vert.x worker thread, as indicated by the following comment:

// CRITICAL: The subscription itself must run on a background thread to avoid blocking
// the Vert.x worker thread. EventConsumer.consumeAll() starts a polling loop that
// blocks in dequeueEventItem(), so we must subscribe from a background thread.
// Use the @Internal executor (not ForkJoinPool.commonPool) to avoid saturation

🔗 [ResultAggregator.java#L121-L125](https://github.com/a2aproject/a2a-java/blob/5031b8faba7b78bbd508a3d21a6b1940ecf19013/server-common/src/main/java/io/a2a/server/tasks/ResultAggregator.java#L121-L125)

However, for requests where blocking=true, it is still expected that the DefaultRequestHandler waits for background task completion and returns the final aggregated result.

Currently, even though cleanupProducer calls bothComplete.join(), the handler does not appear to wait as intended.

🔗 [DefaultRequestHandler.java#L607](https://github.com/a2aproject/a2a-java/blob/5031b8faba7b78bbd508a3d21a6b1940ecf19013/server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java#L607)

## Expected Behavior

When blocking=true, DefaultRequestHandler should:

Wait for all background tasks to complete (including final state aggregation).

Return the final response only after all events have been processed.

## Actual Behavior

The first message is returned immediately.

Remaining events are processed asynchronously in the background.

Behavior is effectively identical to blocking=false.

## Request

Please review the intended behavior for blocking=true requests and confirm whether the current implementation aligns with the design goals.
If not, it may require ensuring that bothComplete.join() properly waits for all background processing before returning the response.

### Relevant log output

```shell

```

### Code of Conduct

kabir:
…cking calls

Fixes race condition where DefaultRequestHandler.onMessageSend() returns before all task events are fully processed and persisted to TaskStore, resulting in incomplete Task objects being returned to clients (missing artifacts, incorrect state).

Root Cause:
- Blocking calls interrupted immediately after first event and returned to client before background event consumption completed
- Agent execution and event processing happened asynchronously in background
- No synchronization to ensure all events were consumed and persisted before returning Task to client

Solution (4-step process):
1. Wait for agent to finish enqueueing events (configurable timeout)
2. Close the queue to signal consumption can complete (breaks dependency)
3. Wait for consumption to finish processing events (configurable timeout)
4. Fetch final task state from TaskStore (has all artifacts and correct state)

This ensures blocking calls return complete Task objects with all artifacts and correct state, including support for fire-and-forget tasks that never emit final state events.

Added unit tests:
- testBlockingFireAndForgetReturnsNonFinalTask: Validates fire-and-forget pattern
- testBlockingCallReturnsCompleteTaskWithArtifacts: Ensures all artifacts included

Fixes #428 🦕

### DIFF
--- a/server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
+++ b/server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
@@ -3,6 +3,7 @@ package io.a2a.server.requesthandlers;
 import static io.a2a.server.util.async.AsyncUtils.convertingProcessor;
 import static io.a2a.server.util.async.AsyncUtils.createTubeConfig;
 import static io.a2a.server.util.async.AsyncUtils.processor;
+import static java.util.concurrent.TimeUnit.*;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -266,7 +267,7 @@ public class DefaultRequestHandler implements RequestHandler {
                     // Step 1: Wait for agent to finish (with configurable timeout)
                     if (agentFuture != null) {
                         try {
-                            agentFuture.get(agentCompletionTimeoutSeconds, java.util.concurrent.TimeUnit.SECONDS);
+                            agentFuture.get(agentCompletionTimeoutSeconds, SECONDS);
                             LOGGER.debug("Agent completed for task {}", taskId);
                         } catch (java.util.concurrent.TimeoutException e) {
                             // Agent still running after timeout - that's fine, events already being processed
@@ -282,16 +283,22 @@ public class DefaultRequestHandler implements RequestHandler {
 
                     // Step 3: Wait for consumption to complete (now that queue is closed)
                     if (etai.consumptionFuture() != null) {
-                        etai.consumptionFuture().get(consumptionCompletionTimeoutSeconds, java.util.concurrent.TimeUnit.SECONDS);
+                        etai.consumptionFuture().get(consumptionCompletionTimeoutSeconds, SECONDS);
                         LOGGER.debug("Consumption completed for task {}", taskId);
                     }
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
-                    LOGGER.warn("Interrupted waiting for task {} completion", taskId, e);
+                    String msg = String.format("Error waiting for task %s completion", taskId);
+                    LOGGER.warn(msg, e);
+                    throw new InternalError(msg);
                 } catch (java.util.concurrent.ExecutionException e) {
-                    LOGGER.warn("Error during task {} execution", taskId, e.getCause());
+                    String msg = String.format("Error during task %s execution", taskId);
+                    LOGGER.warn(msg, e.getCause());
+                    throw new InternalError(msg);
                 } catch (java.util.concurrent.TimeoutException e) {
-                    LOGGER.warn("Timeout waiting for consumption to complete for task {}", taskId);
+                    String msg = String.format("Timeout waiting for consumption to complete for task %s", taskId);
+                    LOGGER.warn(msg, taskId);
+                    throw new InternalError(msg);
                 }
 
                 // Step 4: Fetch the final task state from TaskStore (all events have been processed)


kabir:
@flex-myeonghyeon I moved the work here

~Note that for some reason, GitHub isn't showing the code I have locally, so if you want to check you might have to pull down the branch.~ This was my IDE not behaving. It should be the same now :-)

kabir:
Replaces #430

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -210,7 +214,63 @@ public EventKind onMessageSend(MessageSendParams params, ServerCallContext conte
...
+                        LOGGER.debug("Consumption completed for task {}", taskId);
+                    }
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    LOGGER.warn("Interrupted waiting for task {} completion", taskId, e);
+                } catch (java.util.concurrent.ExecutionException e) {
+                    LOGGER.warn("Error during task {} execution", taskId, e.getCause());
+                } catch (java.util.concurrent.TimeoutException e) {
+                    LOGGER.warn("Timeout waiting for consumption to complete for task {}", taskId);
+                }
```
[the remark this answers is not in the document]
kabir: Fixed

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -63,6 +63,8 @@
 public class DefaultRequestHandler implements RequestHandler {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(DefaultRequestHandler.class);
+    private static final int AGENT_COMPLETION_TIMEOUT_SECONDS = 5;
+    private static final int CONSUMPTION_COMPLETION_TIMEOUT_SECONDS = 2;
```
flex-myeonghyeon: In some cases, a longer wait time might be necessary.
Would it be possible to make this configurable — for example, by passing it as a parameter or injecting it through the DefaultRequestHandler constructor?

kabir: @flex-myeonghyeon Good idea, I've added some MP Config properties

flex-myeonghyeon:
@kabir 
Looks good to me!
Once the snapshot version is deployed and available, I’ll test it in my development environment.

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -93,6 +115,19 @@ public DefaultRequestHandler(AgentExecutor agentExecutor, TaskStore taskStore,
...
 
+    /**
+     * For testing
+     */
+    public static DefaultRequestHandler create(AgentExecutor agentExecutor, TaskStore taskStore,
+                         QueueManager queueManager, PushNotificationConfigStore pushConfigStore,
+                         PushNotificationSender pushSender, Executor executor) {
+        DefaultRequestHandler handler =
+                new DefaultRequestHandler(agentExecutor, taskStore, queueManager, pushConfigStore, pushSender, executor);
+        handler.agentCompletionTimeoutSeconds = 5;
```
ehsavoie, line 126: Shouldn't you pass those values as parameters for testing ?

PR #432, opened 2025-11-03
Backports to 0.3.x CI

kabir:
Just to get CI. We will merge these directly to avoid the annoying squashing


kabir:
Merged manually

PR #453, opened 2025-11-07
WIP

ehsavoie:
Simple TimeStamp extension.


fjuma:
Looks good. Would be good to add some info to the README about this example.

fjuma:
@kabir @ehsavoie Just wondering about the location of this extension. For Python, the corresponding extension is in the a2a-samples project (https://github.com/a2aproject/a2a-samples/tree/main/samples/python/extensions/timestamp), it's not part of the Python SDK repo. We could probably do the same, i.e., add it to samples/java/extensions. WDYT?

ehsavoie:
@fjuma I agree with you  I can move it there

Issue #359, opened 2025-10-20
[Feat]: The ability to list tasks (tasks/list) in addition to the existing get (tasks/get)


fjuma:
See https://github.com/a2aproject/A2A/issues/761

PR #459, opened 2025-11-12
feat: update tasks/list implementation to match A2A 0.4.0 spec

kabir:
- Sort tasks by status.timestamp DESC (most recent first), then ID ASC
  - Add lastUpdatedAfter parameter filtering
  - Update InMemoryTaskStore and JpaDatabaseTaskStore sorting logic
  - Add timestamp denormalized column to JpaTask for efficient querying
  - Update all transport handlers (JSON-RPC, gRPC, REST)

Follows up on #455 and #359

### DIFF
--- a/server-common/src/main/java/io/a2a/server/tasks/InMemoryTaskStore.java
+++ b/server-common/src/main/java/io/a2a/server/tasks/InMemoryTaskStore.java
@@ -47,7 +47,11 @@ public class InMemoryTaskStore implements TaskStore, TaskStateProvider {
                         (task.getStatus() != null &&
                          task.getStatus().timestamp() != null &&
                          task.getStatus().timestamp().toInstant().isAfter(params.lastUpdatedAfter())))
-                .sorted(Comparator.comparing((Task t) -> t.getStatus().timestamp(),
+                .sorted(Comparator.comparing(
+                        (Task t) -> (t.getStatus() != null && t.getStatus().timestamp() != null)
+                                // Truncate to milliseconds for consistency with pageToken precision
+                                ? t.getStatus().timestamp().toInstant().truncatedTo(java.time.temporal.ChronoUnit.MILLIS)
+                                : null,
                         Comparator.nullsLast(Comparator.reverseOrder()))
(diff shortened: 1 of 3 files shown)


server-common/src/main/java/io/a2a/server/tasks/InMemoryTaskStore.java
```diff
@@ -63,21 +69,17 @@ public ListTasksResult list(ListTasksParams params) {
         int pageSize = params.getEffectivePageSize();
         int startIndex = 0;
 
-        // Handle page token (simple cursor: last task ID from previous page)
+        // Handle page token (cursor: task ID from previous page)
+        // Since we're sorted by timestamp DESC then ID ASC, we can't use binary search
+        // Instead, find the task with the matching ID using linear search
         if (params.pageToken() != null && !params.pageToken().isEmpty()) {
```
ehsavoie, line 67: should be in the srteam processing shouldn't it ?

reference/rest/src/main/java/io/a2a/server/rest/quarkus/A2AServerRoutes.java
ehsavoie: valueOf

ehsavoie, line 142: valueOf

ehsavoie, line 137: valueOf

ehsavoie (changes requested):

reference/rest/src/main/java/io/a2a/server/rest/quarkus/A2AServerRoutes.java
kabir: IntelliJ tells me to use parseInt() due to unneccesary unboxing if using valueOf

kabir: IntelliJ tells me to use parseInt() due to unneccesary unboxing if using valueOf

PR #463, opened 2025-11-18
feat: Add telemetry support.

### LINKED ISSUE #388: [Feat] Add telemetry support
The Python SDK provides support for OpenTelemetry Tracing:

https://github.com/a2aproject/a2a-python/blob/main/README.md?plain=1#L53

We should look into this and provide something similar.

ehsavoie:
Adding support for OpenTelemetry

Fixes #388 🦕

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ release.properties
 .flattened-pom.xml
 *.args
 
+#Claude
+CLAUDE.md
+
 # Eclipse
 .project
 .classpath
@@ -19,8 +22,10 @@ bin/
 *.iml
 *.iws
 
-# NetBeans
+# Apache NetBeans
 nb-configuration.xml
+nbactions.xml
+nbproject/
 
 # Visual Studio Code
 .vscode
@@ -44,10 +49,6 @@ nb-configuration.xml
 /.quarkus/cli/plugins/
 # TLS Certificates
 .certs/
-
-#Apache NetBeans
-nbproject/
-nbactions.xml
 
 # Private Claude config
 .claude/
--- a/client/base/src/main/java/io/a2a/client/ClientBuilder.java
+++ b/client/base/src/main/java/io/a2a/client/ClientBuilder.java
@@ -6,20 +6,26 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
+import java.util.ServiceLoader.Provider;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import io.a2a.client.config.ClientConfig;
 import io.a2a.client.transport.spi.ClientTransport;
 import io.a2a.client.transport.spi.ClientTransportConfig;
 import io.a2a.client.transport.spi.ClientTransportConfigBuilder;
 import io.a2a.client.transport.spi.ClientTransportProvider;
+import io.a2a.client.transport.spi.ClientTransportWrapper;
 import io.a2a.spec.A2AClientException;
 import io.a2a.spec.AgentCard;
 import io.a2a.spec.AgentInterface;
 import io.a2a.spec.TransportProtocol;
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Builder for creating instances of {@link Client} to communicate with A2A agents.
@@ -96,6 +102,7 @@ public class ClientBuilder {
 
     private static final Map<String, ClientTransportProvider<? extends ClientTransport, ? extends ClientTransportConfig<?>>> transportProviderRegistry = new HashMap<>();
     private static final Map<Class<? extends ClientTransport>, String> transportProtocolMapping = new HashMap<>();
+    private static final Logger LOGGER = LoggerFactory.getLogger(ClientBuilder.class);
 
     static {
         ServiceLoader<ClientTransportProvider> loader = ServiceLoader.load(ClientTransportProvider.class);
@@ -318,7 +325,7 @@ public class ClientBuilder {
             throw new A2AClientException("Missing required TransportConfig for " + agentInterface.protocolBinding());
         }
 
-        return clientTransportProvider.create(clientTransportConfig, agentCard, agentInterface);
+        return wrap(clientTransportProvider.create(clientTransportConfig, agentCard, agentInterface), clientTransportConfig);
     }
 
     private Map<String, String> getServerPreferredTransports() throws A2AClientException {
@@ -373,10 +380,55 @@ public class ClientBuilder {
         if (transportProtocol == null || transportUrl == null) {
             throw new A2AClientException("No compatible transport found");
         }
-        if (! transportProviderRegistry.containsKey(transportProtocol)) {
+        if (!transportProviderRegistry.containsKey(transportProtocol)) {
             throw new A2AClientException("No client available for " + transportProtocol);
         }
 
         return new AgentInterface(transportProtocol, transportUrl);
+    }
+
+    /**
+     * Wraps the transport with all available transport wrappers discovered via ServiceLoader.
+     * Wrappers are applied in reverse priority order (lowest priority first) to build a stack
+     * where the highest priority wrapper is the outermost layer.
+     *
+     * @param transport the base transport to wrap
+     * @param clientTransportConfig the transport configuration
+     * @return the wrapped transport (or original if no wrappers are available/applicable)
+     */
+    private ClientTransport wrap(ClientTransport transport, ClientTransportConfig<? extends ClientTransport> clientTransportConfig) {
+        ServiceLoader<ClientTransportWrapper> wrapperLoader = ServiceLoader.load(ClientTransportWrapper.class);
+
+        // Collect all wrappers, sort by priority, then reverse for stack application
+        List<ClientTransportWrapper> wrappers = wrapperLoader.stream().map(Provider::get)
+                .sorted()
+                .collect(Collectors.toList());
+
+        if (wrappers.isEmpty()) {
+            LOGGER.debug("No client transport wrappers found via ServiceLoader");
+            return transport;
+        } 
+        LOGGER.debug(wrappers.size() + " client transport wrappers found via ServiceLoader");
+
+        // Reverse to apply lowest priority first (building stack with highest priority outermost)
+        java.util.Collections.reverse(wrappers);
+
+        // Apply wrappers to build stack
+        ClientTransport wrapped = transport;
+        for (ClientTransportWrapper wrapper : wrappers) {
+            try {
+                ClientTransport newWrapped = wrapper.wrap(wrapped, clientTransportConfig);
+                if (newWrapped != wrapped) {
+                    LOGGER.debug("Applied transport wrapper: {} (priority: {})",
+                            wrapper.getClass().getName(), wrapper.priority());
+                }
+                wrapped = newWrapped;
+            } catch (Exception e) {
+                LOGGER.warn("Failed to apply transport wrapper {}: {}",
+                        wrapper.getClass().getName(), e.getMessage(), e);
+            }
+        }
+
+        return wrapped;
     }
 }
--- a/examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldClient.java
+++ b/examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldClient.java
@@ -1,5 +1,8 @@
 package io.a2a.examples.helloworld;
 
+import static io.a2a.extras.opentelemetry.client.OpenTelemetryClientTransportWrapper.OTEL_TRACER_KEY;
+import static io.a2a.extras.opentelemetry.client.propagation.OpenTelemetryClientPropagatorTransportWrapper.OTEL_OPEN_TELEMETRY_KEY;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -11,16 +14,32 @@ import java.util.function.Consumer;
 import io.a2a.A2A;
 
 import io.a2a.client.Client;
+import io.a2a.client.ClientBuilder;
 import io.a2a.client.ClientEvent;
 import io.a2a.client.MessageEvent;
 import io.a2a.client.http.A2ACardResolver;
+import io.a2a.client.transport.grpc.GrpcTransport;
+import io.a2a.client.transport.grpc.GrpcTransportConfigBuilder;
 import io.a2a.client.transport.jsonrpc.JSONRPCTransport;
 import io.a2a.client.transport.jsonrpc.JSONRPCTransportConfig;
+import io.a2a.client.transport.rest.RestTransport;
+import io.a2a.client.transport.rest.RestTransportConfig;
+import io.a2a.client.transport.spi.ClientTransportConfig;
 import io.a2a.jsonrpc.common.json.JsonUtil;
 import io.a2a.spec.AgentCard;
 import io.a2a.spec.Message;
 import io.a2a.spec.Part;
 import io.a2a.spec.TextPart;
+import io.grpc.Channel;
+import io.grpc.ManagedChannelBuilder;
+import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator;
+import io.opentelemetry.context.propagation.ContextPropagators;
+import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
+import java.util.function.Function;
 
 /**
  * A simple example of using the A2A Java SDK to communicate with an A2A server.
@@ -32,13 +51,13 @@ public class HelloWorldClient {
     private static final String MESSAGE_TEXT = "how much is 10 USD in INR?";
 
     public static void main(String[] args) {
+        OpenTelemetrySdk openTelemetrySdk = null;
         try {
-            AgentCard finalAgentCard = null;
             AgentCard publicAgentCard = new A2ACardResolver("http://localhost:9999").getAgentCard();
             System.out.println("Successfully fetched public agent card:");
             System.out.println(JsonUtil.toJson(publicAgentCard));
             System.out.println("Using public agent card for client initialization (default).");
-            finalAgentCard = publicAgentCard;
+            AgentCard finalAgentCard = publicAgentCard;
 
             if (publicAgentCard.capabilities().extendedAgentCard()) {
                 System.out.println("Public card supports authenticated extended card. Attempting to fetch from: " + SERVER_URL + "/ExtendedAgentCard");
@@ -81,20 +100,23 @@ public class HelloWorldClient {
                 messageResponse.completeExceptionally(error);
             };
 
-            Client client = Client
+            if (Boolean.getBoolean("opentelemetry")) {
+                openTelemetrySdk = initOpenTelemetry();
+            }
+
+            ClientBuilder clientBuilder = Client
                     .builder(finalAgentCard)
                     .addConsumers(consumers)
-                    .streamingErrorHandler(streamingErrorHandler)
-                    .withTransport(JSONRPCTransport.class, new JSONRPCTransportConfig())
-                    .build();
+                    .streamingErrorHandler(streamingErrorHandler);
+            configureTransport(clientBuilder, openTelemetrySdk);
+             Client client = clientBuilder.build();
 
             Message message = A2A.toUserMessage(MESSAGE_TEXT); // the message ID will be automatically generated for you
-            
-            System.out.println("Sending message: " + MESSAGE_TEXT);
-            client.sendMessage(message);
-            System.out.println("Message sent successfully. Responses will be handled by the configured consumers.");
-
             try {
+                System.out.println("Sending message: " + MESSAGE_TEXT);
+                client.sendMessage(message);
+                System.out.println("Message sent successfully. Responses will be handled by the configured consumers.");
+
                 String responseText = messageResponse.get();
                 System.out.println("Response: " + responseText);
             } catch (Exception e) {
@@ -103,7 +125,69 @@ public class HelloWorldClient {
         } catch (Exception e) {
             System.err.println("An error occurred: " + e.getMessage());
             e.printStackTrace();
+        } finally {
+            // Ensure OpenTelemetry SDK is properly shut down to export all pending spans
+            if (openTelemetrySdk != null) {
+                System.out.println("Shutting down OpenTelemetry SDK...");
+                openTelemetrySdk.close();
+                System.out.println("OpenTelemetry SDK shutdown complete.");
+            }
         }
     }
 
-} 
+    static OpenTelemetrySdk initOpenTelemetry() {
+        SdkTracerProvider sdkTracerProvider = SdkTracerProvider.builder()
+                .addSpanProcessor(BatchSpanProcessor.builder(
+                        OtlpGrpcSpanExporter.builder()
+                                .setEndpoint("http://localhost:5317")
+                                .build()
+                ).build())
+                .setResource(Resource.getDefault().toBuilder()
+                        .put("service.version", "1.0")
+                        .put("service.name", "helloworld-client")
+                        .build())
+                .build();
+
+        return OpenTelemetrySdk.builder()
+                .setTracerProvider(sdkTracerProvider)
+                .setPropagators(ContextPropagators.create(W3CTraceContextPropagator.getInstance()))
+                .build();
+    }
+    private static void configureTransport(ClientBuilder clientBuilder, OpenTelemetrySdk openTelemetrySdk) {
+        ClientTransportConfig transportConfig;
+        switch(System.getProperty("quarkus.agentcard.protocol", "JSONRPC")) {
+            case "GRPC":
+                Function<String, Channel> channelFactory = url -> {
+                    // Extract "localhost:9999" from "http://localhost:9999"
+                    String target = url.replaceAll("^https?://", "");
+                    return ManagedChannelBuilder.forTarget(target)
+                            .usePlaintext() // No TLS
+                            .build();
+                };
+                transportConfig = new GrpcTransportConfigBuilder().channelFactory(channelFactory).build();
+                updateTransportConfig(transportConfig, openTelemetrySdk);
+                clientBuilder.withTransport(GrpcTransport.class, transportConfig);
+                break;
+            case "HTTP+JSON":
+                transportConfig = new RestTransportConfig();
+                updateTransportConfig(transportConfig, openTelemetrySdk);
+                clientBuilder.withTransport(RestTransport.class, transportConfig);
+                break;
+            case "JSONRPC":
+            default:
+                transportConfig = new JSONRPCTransportConfig();
+                updateTransportConfig(transportConfig, openTelemetrySdk);
+                clientBuilder.withTransport(JSONRPCTransport.class, transportConfig);
+                break;
+        }
+    }
+
+    private static void updateTransportConfig(ClientTransportConfig transportConfig, OpenTelemetrySdk openTelemetrySdk) {
+        if (openTelemetrySdk != null) {
+            Map<String, Object> parameters = new HashMap<>(transportConfig.getParameters());
+            parameters.put(OTEL_TRACER_KEY, openTelemetrySdk.getTracer("helloworld-client"));
+            parameters.put(OTEL_OPEN_TELEMETRY_KEY, openTelemetrySdk);
+            transportConfig.setParameters(parameters);
+        }
+    }
+}
--- a/examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldRunner.java
+++ b/examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldRunner.java
@@ -1,6 +1,16 @@
+
 ///usr/bin/env jbang "$0" "$@" ; exit $?
+
 //DEPS io.github.a2asdk:a2a-java-sdk-client:1.0.0.Alpha2-SNAPSHOT
 //DEPS io.github.a2asdk:a2a-java-sdk-client-transport-jsonrpc:1.0.0.Alpha2-SNAPSHOT
+//DEPS io.github.a2asdk:a2a-java-sdk-client-transport-grpc:1.0.0.Alpha2-SNAPSHOT
+//DEPS io.github.a2asdk:a2a-java-sdk-client-transport-rest:1.0.0.Alpha2-SNAPSHOT
+//DEPS io.github.a2asdk:a2a-java-sdk-opentelemetry-client:1.0.0.Alpha2-SNAPSHOT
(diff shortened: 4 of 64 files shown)


server-common/src/main/java/io/a2a/server/interceptors/Trace.java
```diff
@@ -0,0 +1,20 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import jakarta.interceptor.InterceptorBinding;
+
+@InterceptorBinding 
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD}) 
+@Inherited 
+@Priority(Interceptor.Priority.APPLICATION)
+public @interface Trace {
```
brunobat: Why not use the OpenTelemetry annotation [`@WithSpan`](https://opentelemetry.io/docs/zero-code/java/agent/annotations/#creating-spans-around-methods-with-withspan) instead of creating a new one?

server-common/src/main/java/io/a2a/server/interceptors/AttributeExtractor.java
```diff
@@ -0,0 +1,8 @@
+package io.a2a.server.interceptors;
+
+import java.util.Map;
+
+@FunctionalInterface
+public interface AttributeExtractor {
```
brunobat: There is also https://opentelemetry.io/docs/zero-code/java/agent/annotations/#adding-attributes-to-the-span-with-spanattribute

server-common/src/main/java/io/a2a/server/interceptors/Kind.java
```diff
@@ -0,0 +1,5 @@
+package io.a2a.server.interceptors;
+
+public enum Kind {
```
brunobat: You are duplicating the OpenTelemetry API...

extras/opentelemetry/src/main/java/io/a2a/extras/opentelemetry/OpenTelemetryClientTransport.java
```diff
@@ -0,0 +1,318 @@
...
+                .setSpanKind(SpanKind.CLIENT);
+        spanBuilder.setAttribute("request", request.toString());
+        Span span = spanBuilder.startSpan();
+        try (Scope scope = span.makeCurrent()) {
+            EventKind ret = delegate.sendMessage(request, context);
+            if (ret != null) {
+                span.setAttribute("response", ret.toString());
+                span.setStatus(StatusCode.OK);
+            }
+            span.end();
```
brunobat: Please end the span in a finally block.
Something like:
``` 
try (Scope scope = span.makeCurrent()) {
    EventKind ret = delegate.sendMessage(request, context);
    if (ret != null) {
        span.setAttribute("response", ret.toString());
        span.setStatus(StatusCode.OK);
    }
    return ret;
} catch (Exception ex) {
    span.setStatus(StatusCode.ERROR, ex.getMessage());
    throw ex;
} finally {
    span.end();
}
```

extras/opentelemetry/src/main/java/io/a2a/extras/opentelemetry/OpenTelemetryClientTransport.java
```diff
@@ -0,0 +1,318 @@
...
+    public void sendMessageStreaming(MessageSendParams request, Consumer<StreamingEventKind> eventConsumer, Consumer<Throwable> errorConsumer, ClientCallContext context) throws A2AClientException {
+        SpanBuilder spanBuilder = tracer.spanBuilder("sendMessageStreaming")
+                .setSpanKind(SpanKind.CLIENT);
+        spanBuilder.setAttribute("request", request.toString());
+        Span span = spanBuilder.startSpan();
+        try (Scope scope = span.makeCurrent()) {
+            delegate.sendMessageStreaming(request, new OpenTelemetryEventConsumer(eventConsumer, tracer, span.getSpanContext()),
+                    new OpentTelemetryErrorConsumer(errorConsumer, tracer, span.getSpanContext()), context);
+            span.setStatus(StatusCode.OK);
+            span.end();
```
brunobat: same

extras/opentelemetry/src/main/java/io/a2a/extras/opentelemetry/OpenTelemetryClientTransport.java
```diff
@@ -0,0 +1,318 @@
...
+                .setSpanKind(SpanKind.CLIENT);
+        spanBuilder.setAttribute("request", request.toString());
+        Span span = spanBuilder.startSpan();
+        try (Scope scope = span.makeCurrent()) {
+            Task ret = delegate.getTask(request, context);
+            if (ret != null) {
+                span.setAttribute("response", ret.toString());
+                span.setStatus(StatusCode.OK);
+            }
+            span.end();
```
brunobat: same

examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldClient.java
```diff
@@ -108,4 +115,16 @@ public static void main(String[] args) {
         }
     }
 
-} 
\ No newline at end of file
+    static OpenTelemetry initOpenTelemetry() {
+        SdkTracerProvider sdkTracerProvider = SdkTracerProvider.builder()
```
brunobat, line 139: This will break frameworks that already build an instance of OTel themselves. 
I think you should just inject the OpenTelemetry instance in the code of the project and provide a different module that can provide implementations for the `@WithSpan` interceptor and the OTel sdk, in case the A2A lib is being used standalone, without an earlier OTel provider.

brunobat (commented):
I think you need to decide if you are creating your own instrumentation annotations and API or if you just reuse the ones provided by OTel.

Also, the attribute names, should follow these semantic conventions, as much as possible: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-agent-spans/

server-common/src/main/java/io/a2a/server/interceptors/Trace.java
ehsavoie: because then I would have to have opentelemetry as a direct dependency

brunobat: you can depend only on `opentelemetry-api`

examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldClient.java
ehsavoie: this is only an example.How should I inject the code in the client part without a CDI container ?

brunobat: You can have a new module that creates this... I mention this later.

brunobat: One thing is sure, as soon as that code hits something like Quarkus, the startup will abort with a failure because there is already an OTel instance created at static time.

brunobat: this is an example. that's fine

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -37,7 +43,8 @@ public class ClientBuilder {
     private final AgentCard agentCard;
 
     private final List<BiConsumer<ClientEvent, AgentCard>> consumers = new ArrayList<>();
-    private @Nullable Consumer<Throwable> streamErrorHandler;
+    private @Nullable
+    Consumer<Throwable> streamErrorHandler;
```
kabir: This looks a bit strange :-)

The original looks better, but I think 

````
@Nullable
private Consumer<Throwable> streamErrorHandler;
````
would look better. But I see we have `private @Nullable ....` 'everywhere' :-)

File: extras/opentelemetry/src/main/java/io/a2a/extras/opentelemetry/SpanInterceptor.java [hunk 9]
```diff
@@ -0,0 +1,93 @@
...
+                .getMethod()
+                .getAnnotation(Trace.class)
+                .kind();
+        Class<? extends Supplier<Function<io.a2a.server.interceptors.InvocationContext, Map<String, String>>>> extractorClass
+                = jakartaContext.getMethod()
+                        .getAnnotation(Trace.class)
+                        .extractor();
+
+        String name = jakartaContext.getTarget().getClass().getName();
+        if (name != null && name.endsWith("_Subclass")) {
```
kabir: name cannot be null

server-common/src/main/java/io/a2a/server/requesthandlers/RequestHandlerAttributeExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+                     "setPushNotificationConfig",
+                     "onGetTask",
+                     "listPushNotificationConfig",
+                     "deletePushNotificationConfig",
+                     "onListTasks" -> {
+                    ServerCallContext context = (ServerCallContext) parameters[1];
+                    return Map.of("request", parameters[0].toString(), "extensions", context.getActivatedExtensions().stream().collect(Collectors.joining(",")), "a2a.method", (String) context.getState().get("method"));
+                }
+                default -> {
+                    return Collections.emptyMap();
```
kabir: Maybe we could log a warning here. Or if the intent is that all DefaultRequestHandler methods are handled, we could throw an error?

transport/grpc/src/main/java/io/a2a/transport/grpc/handler/GrpcAttributeExtractor.java
```diff
@@ -0,0 +1,39 @@
...
+                     "getTaskPushNotificationConfig",
+                     "listTaskPushNotificationConfig",
+                     "sendStreamingMessage",
+                     "taskSubscription",
+                     "deleteTaskPushNotificationConfig" -> {
+                    Context currentContext = Context.current();
+                    return Map.of("request", parameters[0].toString(), "extensions", GrpcContextKeys.EXTENSIONS_HEADER_KEY.get(), "a2a.method", GrpcContextKeys.GRPC_METHOD_NAME_KEY.get(currentContext));
+                }
+                default -> {
+                    return Collections.emptyMap();
```
kabir: log warn/throw error as above

transport/jsonrpc/src/main/java/io/a2a/transport/jsonrpc/handler/JSONRPCAttributeExtractor.java
```diff
@@ -0,0 +1,41 @@
...
+                     "setPushNotificationConfig",
+                     "onGetTask",
+                     "listPushNotificationConfig",
+                     "deletePushNotificationConfig",
+                     "onListTasks" -> {
+                    ServerCallContext context = (ServerCallContext) parameters[1];
+                    return Map.of("body", parameters[0].toString(), "extensions", context.getActivatedExtensions().stream().collect(Collectors.joining(",")), "a2a.method", (String) context.getState().get(METHOD_NAME_KEY));
+                }
+                default -> {
+                    return Collections.emptyMap();
```
kabir: warn/error

transport/rest/src/main/java/io/a2a/transport/rest/handler/RestAttributeExtractor.java
```diff
@@ -0,0 +1,56 @@
...
+                     "deleteTaskPushNotificationConfiguration" -> {
+                    ServerCallContext context = (ServerCallContext) parameters[2];
+                    return Map.of("taskId", (String) parameters[0], "configId", (String) parameters[1], "extensions", context.getActivatedExtensions().stream().collect(Collectors.joining(",")), "a2a.method", (String) context.getState().get(METHOD_NAME_KEY));
+                }
+                case "listTasks" -> {
+                    ServerCallContext context = (ServerCallContext) parameters[6];
+                    return Map.of("contextId", (String) parameters[0], "status", (String) parameters[1], "extensions", context.getActivatedExtensions().stream().collect(Collectors.joining(",")), "a2a.method", (String) context.getState().get(METHOD_NAME_KEY));
+                }
+                default -> {
+                    return Collections.emptyMap();
```
kabir: warn/error

kabir (commented):
A few things which may or may not be needed :-)

kabir (dismissed):
Actually, @ehsavoie it looks like you are adding Otel tracing to the example. But I don't know how to enable this etc.
Can you update the README with instructions?

Maybe we could also make this example testable in the CI, like we do for the cloud-deployment one? 

Also, it would be good to have some tests in the extras module

examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldRunner.java
```diff
@@ -1,6 +1,11 @@
 ///usr/bin/env jbang "$0" "$@" ; exit $?
+
 //DEPS io.github.a2asdk:a2a-java-sdk-client:1.0.0.Alpha1-SNAPSHOT
 //DEPS io.github.a2asdk:a2a-java-sdk-client-transport-jsonrpc:1.0.0.Alpha1-SNAPSHOT
+//DEPS io.github.a2asdk:a2a-java-sdk-opentelemetry:1.0.0.Alpha1-SNAPSHOT
+//DEPS io.opentelemetry:opentelemetry-sdk:1.44.1
```
kabir: The validate-jbang-versions.groovy script stuff needs updating to make sure this version doesn't go out of sync with the poms

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -108,7 +115,8 @@ public class ClientBuilder {
...
```
kabir: This still looks weird :-)

kabir (commented):
@ehsavoie 2 small things.

If the JBang validator is tricky, open an issue and link it here

extras/opentelemetry/src/main/java/io/a2a/extras/opentelemetry/OpenTelemetryClientTransport.java
```diff
@@ -0,0 +1,322 @@
...
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Scope;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.jspecify.annotations.Nullable;
+
+public class OpenTelemetryClientTransport implements ClientTransport {
```
brunobat: How are you handling the OTel context propagation?
We need to account for parent spans and also propagate state using the appropriate headers during transport.
If we don't do this, spans will appear detached from each-other.

brunobat (changes requested):

extras/opentelemetry/src/main/java/io/a2a/extras/opentelemetry/OpenTelemetryClientTransport.java
ehsavoie: @brunobat can you point me to which headers i should use to propagate ?

ehsavoie: I think I will also need to avoid tracing the full request / response

brunobat: > @brunobat can you point me to which headers i should use to propagate ?

Please take a look at this Quarkus doc: https://quarkus.io/version/main/guides/opentelemetry-tracing#manual-context-propagation

brunobat: > I think I will also need to avoid tracing the full request / response

Yes, that will generate an astronomical amount of telemetry data. Can you make it configurable?... I imagine it can be useful in many cases.

ehsavoie: Done

client/base/src/main/java/io/a2a/client/ClientBuilder.java
```diff
@@ -373,10 +380,50 @@ private AgentInterface findBestClientTransport() throws A2AClientException {
...
+     * Wrappers are applied in priority order (highest priority first).
+     *
+     * @param transport the base transport to wrap
+     * @param clientTransportConfig the transport configuration
+     * @return the wrapped transport (or original if no wrappers are available/applicable)
+     */
+    private ClientTransport wrap(ClientTransport transport, ClientTransportConfig<? extends ClientTransport> clientTransportConfig) {
+        ServiceLoader<ClientTransportWrapper> wrapperLoader = ServiceLoader.load(ClientTransportWrapper.class);
+
+        // Collect all wrappers and sort by natural order (uses Comparable implementation)
```
jmesnil: is that correct? Their comparison is done by their priority.

extras/opentelemetry/common/src/main/java/io/a2a/extras/opentelemetry/A2AObservabilityNames.java
```diff
@@ -0,0 +1,20 @@
+package io.a2a.server.interceptors;
+
+
+public interface A2AObservabilityNames {
+    String ERROR_TYPE = "error.type";
+
+    String GENAI_PREFIX = "gen_ai.agent.a2a";
+    String GENAI_CONFIG_ID = GENAI_PREFIX + ".config_id";
+    String GENAI_CONTEXT_ID = GENAI_PREFIX + ".context_id"; //gen_ai.conversation.id
```
jmesnil, line 12: what does the comment mean?

server-common/src/main/java/io/a2a/server/interceptors/A2AObservabilityNames.java
```diff
@@ -0,0 +1,20 @@
...
+
+public interface A2AObservabilityNames {
+    String ERROR_TYPE = "error.type";
+
+    String GENAI_PREFIX = "gen_ai.agent.a2a";
+    String GENAI_CONFIG_ID = GENAI_PREFIX + ".config_id";
+    String GENAI_CONTEXT_ID = GENAI_PREFIX + ".context_id"; //gen_ai.conversation.id
+    String GENAI_EXTENSIONS = GENAI_PREFIX + ".extensions";
+    String GENAI_MESSAGE_ID = GENAI_PREFIX + ".message_id";
+    String GENAI_OPERATION_NAME = "gen_ai.agent.operation.name";
```
jmesnil: why is that one no using the prefix?

server-common/src/main/java/io/a2a/server/requesthandlers/RequestHandlerAttributeExtractor.java
```diff
@@ -0,0 +1,101 @@
...
+                case "onMessageSend", "onMessageSendStream" -> {
+                    Map<String, String> attributes = processRequest(parameters);
+                    if (parameters[0] != null && parameters[0] instanceof MessageSendParams messageSendParams) {
+                        processMessage(attributes, messageSendParams.message());
+                    }
+                    return attributes;
+                }
+                case "onCancelTask",
+                     "onResubscribeToTask",
+                     "getPushNotificationConfig",
```
jmesnil: not related to your PR but the names of the methods lack consistency

transport/grpc/src/main/java/io/a2a/transport/grpc/context/GrpcContextKeys.java
```diff
@@ -24,28 +28,47 @@ public final class GrpcContextKeys {
...
+            Context.key("method");
+
     /**
      * Context key for storing the peer information.
      * Provides access to client connection details.
      */
     public static final Context.Key<String> PEER_INFO_KEY = 
         Context.key("grpc-peer-info");
 
+    public static final Map<String, String> METHOD_MAPPING = Map.of(
```
jmesnil, line 60: aren't the keys and the values the same?

transport/rest/src/main/java/io/a2a/transport/rest/handler/RestAttributeExtractor.java
```diff
@@ -0,0 +1,121 @@
...
+                     "deleteTaskPushNotificationConfiguration" -> {
+                    ServerCallContext context = (ServerCallContext) parameters[3];
+                    Map<String, String> result = new HashMap<>();
+                    putIfNotNull(result, GENAI_TASK_ID, (String) parameters[0]);
+                    putIfNotNull(result, GENAI_CONFIG_ID, (String) parameters[1]);
+                    result.putAll(processServerCallContext(context));
+                    return result;
+                }
+                case "listTasks" -> {
+                    ServerCallContext context = (ServerCallContext) parameters[7];
```
jmesnil: this is very brittle to have magic number to get the serverCallContext parameter (there is no way I'll remember to uppdate this class when I add a parameter to one of these methods...)
There is no safeguard that the given parameter is an instance of  ServerCallContext.
Maybe we should update the code and make the server call context *always* the first parameter

examples/helloworld/server/pom.xml
```diff
@@ -25,9 +25,24 @@
             <artifactId>a2a-java-sdk-reference-jsonrpc</artifactId>
         </dependency>
         <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-resteasy-jackson</artifactId>
-            <scope>provided</scope>
+            <groupId>io.github.a2asdk</groupId>
+            <artifactId>a2a-java-sdk-transport-jsonrpc</artifactId>
```
jmesnil: do you need the transport dependencies? (in addition to the reference dependencies)

client/base/src/main/java/io/a2a/client/ClientBuilder.java
ehsavoie: it brings confusion, so I'll remove that

extras/opentelemetry/common/src/main/java/io/a2a/extras/opentelemetry/A2AObservabilityNames.java
ehsavoie: I put here some https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/ entries that might be used

server-common/src/main/java/io/a2a/server/interceptors/A2AObservabilityNames.java
ehsavoie: it uses https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/ value, but I will change that

transport/rest/src/main/java/io/a2a/transport/rest/handler/RestAttributeExtractor.java
ehsavoie: +1

server-common/src/main/java/io/a2a/server/requesthandlers/RequestHandlerAttributeExtractor.java
ehsavoie: yes

transport/grpc/src/main/java/io/a2a/transport/grpc/context/GrpcContextKeys.java
ehsavoie: No, CreateTaskPushNotification is SET_TASK_PUSH_NOTIFICATION_CONFIG_METHOD. Maybe we should align those values

jmesnil: that will be addressed in https://github.com/a2aproject/a2a-java/pull/614/changes#diff-35d00dd67929b417306dc17ceb13343fa70930411ae51dd9d91c056952716b8d

ehsavoie: Yes, we will remove it once that's done

transport/rest/src/main/java/io/a2a/transport/rest/handler/RestAttributeExtractor.java
ehsavoie: fixed

kabir (approved):

examples/helloworld/server/src/main/java/io/a2a/examples/helloworld/AgentCardProducer.java
```diff
@@ -9,24 +9,30 @@
...
 import io.a2a.spec.AgentSkill;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 
 @ApplicationScoped
 public class AgentCardProducer {
 
+    @ConfigProperty(name = "quarkus.agentcard.protocol", defaultValue="JSONRPC")
+    String protocol;
+
+    @Trace
```
brunobat: Sorry to bring this up again but I continue to be concerned with this annotation and interceptor. 
Mind that Quarkus already has the interceptor for `@WithSpan`, which is defined by OpenTelemetry and is also part of Microprofile Telemetry. Most likely they will clash and people will be confused on which one they should use. 
This should be valid for Wildfly and Spring as well.

If you want to avoid depending on the OpenTelemetry API, the best way is to create an SPI that you can implement with OpenTelemetry.

I think you need more complex IT cases where people already use existing instrumentation in order for you to understand the implications of this new instrumentation. 

If this needs to work with Spring that interceptor impl will not work there.

We also need to discuss how this context propagation will work with existing transport layers... Ex. Many HTTP clients already send OTel context data by default.

This is why We need minimal APPs doing some work where we can validate the tracing behaviour.

brunobat (changes requested):

examples/helloworld/server/src/main/java/io/a2a/examples/helloworld/AgentCardProducer.java
brunobat: I meet with @ehsavoie and decided that the propagation part should be on an independent optional module.
No need for `@Traced`. 
There will be a more complete IT test to see how the A2A tracing integrates with existing spans in the apps.

ehsavoie: I've rewritten the whole PR

extras/opentelemetry/integration-tests/src/test/java/io/a2a/extras/opentelemetry/it/OpenTelemetryIT.java
```diff
@@ -0,0 +1,325 @@
...
+import java.util.Map;
+
+/**
+ * Integration test for OpenTelemetry tracing in A2A SDK.
+ * <p>
+ * This test uses the Client API and TestUtilsBean to validate that
+ * OpenTelemetry spans are created correctly for A2A operations.
+ */
+@QuarkusIntegrationTest
+class OpenTelemetryIT extends OpenTelemetryTest {
```
brunobat: Why do you perform all these tests only in native mode?

ehsavoie: What do you mean by native ?

ehsavoie: @brunobat Should be fixed

brunobat: Test classes annotated with `@QuarkusIntegrationTest` will only run in native mode. 
Now you don't have any tests running on native. Usually you have a test class running with `@QuarkusTest` and the another one called *IT annotated with `@QuarkusIntegrationTest` that will run Quarkus native mode tests with graalVM

extras/opentelemetry/client-propagation/src/main/java/io/a2a/extras/opentelemetry/client/propagation/OpenTelemetryClientPropagatorTransportWapper.java
```diff
@@ -0,0 +1,49 @@
...
+ * ClientTransportConfig config = new JSONRPCTransportConfig();
+ * config.setParameters(Map.of(
+ *     OpenTelemetryClientTransportFactory.OTEL_TRACER_KEY,
+ *     openTelemetry.getTracer("my-service"),
+ *     OpenTelemetryClientTransportFactory.OTEL_OPEN_TELEMETRY_KEY,
+ *     openTelemetry
+ * ));
+ * }</pre>
+ */
+public class OpenTelemetryClientPropagatorTransportWapper implements ClientTransportWrapper {
```
kabir: s/Wapper/Wrapper

.gitignore
```diff
@@ -54,3 +58,4 @@ nbactions.xml
 .serena/
 .bob/
 claudedocs
+.bob/
```
kabir: Already on line 59

.gitignore
```diff
@@ -21,6 +24,7 @@ bin/
 
 # NetBeans
 nb-configuration.xml
+nbactions.xml
```
kabir, line 27: There is another NetBeans section further down, which also has this file

spec/src/main/java/io/a2a/spec/Message.java
```diff
@@ -95,6 +95,11 @@ public static Builder builder(Message message) {
         return new Builder(message);
     }
 
+    @Override
+    public String toString() {
```
kabir: Concern:
  - Records auto-generate toString() - explicitly overriding changes public API behavior
  - Used for span attributes when extractRequest/extractResponse flags enabled
  - Consider if this is the right approach for observability data extraction

  Recommendation: Document this as intentional API change, or use separate utility method for telemetry formatting

spec/src/main/java/io/a2a/spec/Task.java
```diff
@@ -98,6 +98,11 @@ public static Builder builder(Task task) {
         return new Builder(task);
     }
 
+    @Override
+    public String toString() {
```
kabir: Same as for Message

extras/opentelemetry/client-propagation/src/main/java/io/a2a/extras/opentelemetry/client/propagation/OpenTelemetryClientPropagatorTransportWapper.java
```diff
@@ -0,0 +1,49 @@
+package io.a2a.extras.opentelemetry.client.propagation;
+
+import io.a2a.client.transport.spi.ClientTransport;
+import io.a2a.client.transport.spi.ClientTransportConfig;
+import io.a2a.client.transport.spi.ClientTransportWrapper;
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.trace.Tracer;
+
+/**
+ * OpenTelemetry client transport wrapper that adds opentelemtry propagation to A2A client calls.
```
kabir: "adds opentelemetry"

kabir (changes requested):

spec/src/main/java/io/a2a/spec/Message.java
ehsavoie: Not used, so removed

extras/opentelemetry/client-propagation/src/main/java/io/a2a/extras/opentelemetry/client/propagation/OpenTelemetryClientPropagatorTransportWapper.java
ehsavoie: fixed

PR #484, opened 2025-11-24
feat!: Updating to newest proto to align with 1.0 Draft specifications

ehsavoie:
[Feat]: Use protobuf as the serialization mechanism


client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransport.java
```diff
@@ -146,6 +145,7 @@ public Task getTask(TaskQueryParams taskQueryParams, @Nullable ClientCallContext
                 throw RestErrorMapper.mapRestError(response);
             }
             String httpResponseBody = response.body();
+            System.out.println("********************************************* " + httpResponseBody);
```
kabir: Remove this :-)

spec-grpc/pom.xml
```diff
@@ -72,75 +75,34 @@
...
-                <groupId>kr.motd.maven</groupId>
-                <artifactId>os-maven-plugin</artifactId>
-                <version>${os-maven-plugin.version}</version>
-            </extension>
-        </extensions>
+        <plugins>
+            <!-- Protocol Buffers Compiler Plugin -->
+            <plugin>
+                <groupId>io.github.ascopes</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
```
kabir, line 94: Is not using the old profile intentional? If so, I think there is a README mentioning the profile which will need removing

fjuma:
It might be good to add some info in the PR description or commit message to indicate the commit hash that was used when copying the .proto file to this repo.

spec-grpc/pom.xml
ehsavoie: Yes because this tool allows me to define both versions (grpc and proto-java) thus we can follow the grpc feature pack more easily

ehsavoie: Also it is not a profile but a  -Dskip.protobuf.generate=false ;)

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransport.java
```diff
@@ -88,7 +76,7 @@ public JSONRPCTransport(String agentUrl) {
     }
 
     public JSONRPCTransport(AgentCard agentCard) {
-        this(null, agentCard, agentCard.url(), null);
+        this(null, agentCard, agentCard.supportedInterfaces().get(0).url(), null);
```
jmesnil: can supportedInterfaces() be empty?

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransport.java
```diff
@@ -159,18 +135,12 @@ public void sendMessageStreaming(MessageSendParams request, Consumer<StreamingEv
...
-
-        PayloadAndHeaders payloadAndHeaders = applyInterceptors(GetTaskRequest.METHOD, getTaskRequest,
+        PayloadAndHeaders payloadAndHeaders = applyInterceptors(io.a2a.spec.GetTaskRequest.METHOD, ProtoUtils.ToProto.getTaskRequest(request),
                 agentCard, context);
 
         try {
-            String httpResponseBody = sendPostRequest(payloadAndHeaders);
-            GetTaskResponse response = unmarshalResponse(httpResponseBody, GET_TASK_RESPONSE_REFERENCE);
+            String httpResponseBody = sendPostRequest(payloadAndHeaders, io.a2a.spec.GetTaskRequest.METHOD);
+            GetTaskResponse response = (GetTaskResponse) unmarshalResponse(httpResponseBody, io.a2a.spec.GetTaskRequest.METHOD);
```
jmesnil: could `unmarshalResponse` be type-safe instead of having to cast the result?

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
```diff
@@ -288,110 +302,104 @@ public class JsonMessages {
...
-                  "kind": "text",
-                  "text": "tell me a joke"
-                 },
-                 {
-                  "kind": "file",
-                  "file": {
-                     "uri": "file:///path/to/file.txt",
-                     "mimeType": "text/plain"
+              "jsonrpc":"2.0",
+              "result":{
```
jmesnil, line 320: is it correct? Shouldn't the result be a JSON object with a `task` field (instead of directly the `task`)?

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
```diff
@@ -288,110 +302,104 @@ public class JsonMessages {
...
-              "id": "de38c76d-d54c-436c-8b9f-4c2703648d64",
-              "contextId": "c295ea44-7543-4f78-b524-7a38915ad6e4",
-              "status": {
-               "state": "canceled"
-              },
-              "metadata": {},
-              "kind" : "task"
-             }
+              "jsonrpc":"2.0",
+              "result":{
```
jmesnil, line 380: idem (I'd be expecting a `task` field)

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
```diff
@@ -764,6 +808,6 @@ public class JsonMessages {
                   }
                 ],
                 "supportsAuthenticatedExtendedCard": true,
-                "protocolVersion": "0.2.5"
+                "protocolVersion": "0.4.0"
```
jmesnil: might be updated to `1.0`?

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonStreamingMessages.java
```diff
@@ -26,16 +27,16 @@ public class JsonStreamingMessages {
...
-                    "role": "agent",
-                    "messageId": "msg-123",
-                    "contextId": "context-456",
-                    "parts": [
-                      {
-                        "kind": "text",
-                        "text": "Hello, world!"
-                      }
-                    ]
+                    "msg": {
```
jmesnil: s/msg/message

client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransport.java
```diff
@@ -62,7 +61,7 @@ public class RestTransport implements ClientTransport {
     private boolean needsExtendedCard = false;
 
     public RestTransport(AgentCard agentCard) {
-        this(null, agentCard, agentCard.url(), null);
+        this(null, agentCard, agentCard.supportedInterfaces().get(0).url(), null);
```
jmesnil: after seeing so many `agentCard.supportedInterfaces().get(0).url()`, I wonder if we should have a `agentCard.favoriteInterfaceUrl()` ;)

reference/jsonrpc/src/main/java/io/a2a/server/apps/quarkus/A2AServerRoutes.java
```diff
@@ -331,9 +383,13 @@ public Buffer apply(Object o) {
                         ReactiveRoutes.ServerSentEvent<?> ev = (ReactiveRoutes.ServerSentEvent<?>) o;
                         long id = ev.id() != -1 ? ev.id() : count.getAndIncrement();
                         String e = ev.event() == null ? "" : "event: " + ev.event() + "\n";
-                        return Buffer.buffer(e + "data: " + Utils.toJsonString(ev.data()) + "\nid: " + id + "\n\n");
+                        String data = serializeResponse((JSONRPCResponse<?>) ev.data());
+                        System.out.println("Sending message " + data);
```
jmesnil: to remove

reference/jsonrpc/src/main/java/io/a2a/server/apps/quarkus/A2AServerRoutes.java
```diff
@@ -331,9 +383,13 @@ public Buffer apply(Object o) {
...
+                        return Buffer.buffer(e + "data: " + data + "\nid: " + id + "\n\n");
                     }
-                    return Buffer.buffer("data: " + Utils.toJsonString(o) + "\nid: " + count.getAndIncrement() + "\n\n");
+                    String data = serializeResponse((JSONRPCResponse<?>) o);
+                    System.out.println("Sending message " + data);
```
jmesnil: idem

spec-grpc/src/main/java/io/a2a/grpc/mapper/DataPartMapper.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+@Mapper(config = A2AProtoMapperConfig.class, uses = {A2ACommonFieldMapper.class})
+public interface DataPartMapper {
+
+    DataPartMapper INSTANCE = A2AMappers.getMapper(DataPartMapper.class);
+
+    /**
+     * Converts domain DataPart to proto DataPart.
+     * Uses CommonFieldMapper for Map → Struct conversion.
+     * Metadata is ignored (not part of proto definition).
```
jmesnil, line 25: does that mean that there can't be metadata attached to a data part?

spec-grpc/pom.xml
```diff
@@ -69,78 +79,42 @@
...
+                <version>4.0.2</version>
+                <configuration>
+                    <skip>${skip.protobuf.generate}</skip>
```
jmesnil, line 97: Why did you remove the profile?

spec/src/main/java/io/a2a/spec/InvalidAgentResponseError.java
```diff
@@ -44,7 +44,7 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class InvalidAgentResponseError extends JSONRPCError {
 
-    public final static Integer DEFAULT_CODE = -32006;
+    public final static int DEFAULT_CODE = -32006;
```
jmesnil: can we get the code from the `.proto` file or somewhere else?

Or at least put all the error code in a single file to simplify their maintenance...

spec/src/main/java/io/a2a/spec/Message.java
```diff
@@ -129,6 +129,15 @@ public String getKind() {
         return kind;
     }
 
+    /**
+     * Creates a new Builder for constructing Message instances.
+     *
+     * @return a new Message.Builder instance
+     */
+    public static Builder builder() {
```
jmesnil, line 137: I much prefer

```
Message m = Message.builder()
  [...]
  .build();
```

to 

```
Message m = new Message.Builder()
  [...]
  .build();
```

should we deprecate the constructor (or make it private)?

spec/src/main/java/io/a2a/spec/Message.java
```diff
@@ -129,6 +129,15 @@ public String getKind() {
         return kind;
     }
 
+    /**
```
jmesnil, line 132: Could we use records to represent the immutable domain objects?

spec-grpc/pom.xml
ehsavoie: because I replaced the plugin and the new one could be skipped

spec/src/main/java/io/a2a/spec/Message.java
ehsavoie: I didn't want to change anything in the spec module

spec/src/main/java/io/a2a/spec/Message.java
ehsavoie: I didn't want to change anything in the spec module

spec/src/main/java/io/a2a/spec/InvalidAgentResponseError.java
ehsavoie: I didn't want to change anything in the spec module but that makes sense

spec-grpc/src/main/java/io/a2a/grpc/mapper/DataPartMapper.java
ehsavoie: @kabir maybe you can answer there

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
ehsavoie: Well I'm not sure about the 'current' version

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
ehsavoie: I think this should require validation in the specs BUT it was like that for rest and since the type is known in advance the task field wouldn't bring any added value (except readability)

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
ehsavoie: I think this should require validation in the specs BUT it was like that for rest and since the type is known in advance the task field wouldn't bring any added value (except readability)

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
jmesnil: and consistency :)

spec-grpc/pom.xml
jmesnil: I see. I prefer to use Maven profiles to control that kind of behaviour but up to you

spec/src/main/java/io/a2a/spec/InvalidAgentResponseError.java
jmesnil: +1, we can keep the `DEFAULT_CODE` so we don't break anything but move the actual values to a single class (eg `spec.ErrorCodes` to simplify the maintenance and checking with https://a2a-protocol.org/latest/specification/#54-error-code-mappings)

jmesnil: and maybe put the mapping to the HTTP status from https://github.com/ehsavoie/a2a-java/blob/058d607975fe02bbb2cf48d7c51b842b0cbeea7d/transport/rest/src/main/java/io/a2a/transport/rest/handler/RestHandler.java#L371 there too?

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
jmesnil: I agree with you that the spec does not clearly specify the returned payload but this would be consistent with https://a2a-protocol.org/latest/specification/#941-sendmessage

ehsavoie: I disagree as the returned payload in that case can be a task or a message while here it is always a task

jmesnil: +1 to open an issue on [A2A](https://github.com/a2aproject/A2A) to clarify this.

spec-grpc/src/main/java/io/a2a/grpc/mapper/DataPartMapper.java
jmesnil: So it seems that the `DataPart`message is protobuf *does not* have metadata according to https://github.com/a2aproject/A2A/blob/b6ef9eec558c877fb69024df090a8bb63c542a1c/specification/grpc/a2a.proto#L228C9-L228C17.
The spec does not mention them either: https://a2a-protocol.org/latest/specification/#418-datapart

So is this something that was present in a previous spec but is no longer relevant?
Now it seems that metadata can only be attached to artifacts and not to their parts.

spec/src/main/java/io/a2a/spec/Message.java
jmesnil: Are you planning to remove the JSON serialisation with Jackson in this PR or in a follow-up?

ehsavoie: I was planning in a follow up to avoid making this big bang pr a bigger bang ;)

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransport.java
ehsavoie: @kabir ^^

jmesnil: According to https://a2a-protocol.org/latest/definitions/, it can be empty:

```
message AgentCard {
  [...]
  // Ordered list of supported interfaces. First entry is preferred.
  repeated AgentInterface supported_interfaces = 19;
}
```

`repeated` means 0 or more

kabir: @ehsavoie I guess if null/empty we should throw some kind of exception

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
kabir: We're aiming for 1.0 (or 1.0.0) so we can update to that

client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransport.java
kabir: @ehsavoie I think we should keep the AgentCard class 'clean', but one possibility could be to add a convenience method to the A2A class perhaps

spec-grpc/src/main/java/io/a2a/grpc/mapper/DataPartMapper.java
kabir: @ehsavoie @jmesnil Yeah, I remember seeing some issues and discussions about metadata being present in the other protocols but not gRPC. 
Maybe we can open a follow-up issue to clean that up? If you agree, please link the issue here so we track it (and add the P0 label)

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
kabir: https://a2a-protocol.org/latest/definitions/#protobuf says:

````
// Get the current state of a task from the agent.
  rpc GetTask(GetTaskRequest) returns (Task) {
    option (google.api.http) = {
      get: "/v1/{name=tasks/*}"
    };
    option (google.api.method_signature) = "name";
  }
````
So I think Emmanuel is right to put the Task directly into the result field of the JSON-RPC wrapper

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
kabir: Again the spec says:
````
 // Cancel a task.
  rpc CancelTask(CancelTaskRequest) returns (Task) {
    option (google.api.http) = {
      post: "/v1/{name=tasks/*}:cancel"
      body: "*"
    };
  }
  ````
  It looks correct to me to put the task directly into the `result` of the JSON-RPC wrapper

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
kabir: https://a2a-protocol.org/latest/specification/#441-agentcard says the following, so we should use `1.0`

Screenshot 2025-12-02 at 14 58 54

client/transport/jsonrpc/src/test/java/io/a2a/client/transport/jsonrpc/JsonMessages.java
jmesnil: OK.
The spec lists the affected types in https://a2a-protocol.org/latest/specification/#a21-breaking-change-kind-discriminator-removed and `task` would be affected only when it is a result of a Streaming Event Types. Not when it is a directly result (such as for `GetTask` and `CancelTask`)

spec-grpc/src/main/java/io/a2a/grpc/mapper/DataPartMapper.java
jmesnil: That'd make sense to me to be able to attach metadata to parts (eg for a file, it would allow to put its creation/update date which would not make sense if it was at the artifact level)

kabir: I created https://github.com/a2aproject/a2a-java/issues/497 to track this, and made it a P0 issue in https://github.com/orgs/a2aproject/projects/4. I don't think this needs to block the merge

spec/src/main/java/io/a2a/spec/Message.java
kabir: As a bit of context, we used to have just the constructor variant to create these. The static `builder()` methods were added to some classes in order to clarify things for the MapStruct mappers.

I agree that `Message.builder()` is nicer, and if we have that everywhere, it makes things more consistent. 
@jmesnil feel free to open an issue. But I think for now we should just deprecate the public constructors and add the static builder methods. Then at a later date we can make the constructors private.

My reasoning is that we're already breaking a lot of stuff with this spec update. Then again maybe breaking more could be ok

jmesnil: I've opened https://github.com/a2aproject/a2a-java/issues/498 to tackle this later.

client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransport.java
jmesnil: +1 for an helper method in `A2A`.

spec/src/main/java/io/a2a/spec/Message.java
kabir: Added a P0 issue https://github.com/a2aproject/a2a-java/issues/499 to https://github.com/orgs/a2aproject/projects/4/views/2

kabir: Ah there was already https://github.com/a2aproject/a2a-java/issues/487. I'll close #499

client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransport.java
ehsavoie: This is only used in tests

spec/src/main/java/io/a2a/spec/InvalidAgentResponseError.java
ehsavoie: I'm not so sure as JSON-RPC doesn't use HTTP Status code, only the rest handler does

jmesnil: If the JSON-RPC error codes have leaked into the domain classes, I don't see why it couldn't also provide the REST mapping?

client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransport.java
ehsavoie: I added a method to Utils to manage that

spec/src/main/java/io/a2a/spec/InvalidAgentResponseError.java
ehsavoie: I'm not sure I'm following you, if we change the status of the response on the json-rpc client side that means we have to handle the error processing on the json-rpc client in a different way.
JSON-RPC is the first supported protocol and as such we refer to it for error codes, method names.

ehsavoie: @jmesnil I think this is not something 'new', so maybe we should address that as part of a new issue

jmesnil: That's minor. I'm just pointing out that the JSON-RPC codes have leaked into the "domain" classes (in `spec`) and are not in the transport layer for JSON-RPC.
That's fine.
My point was that if the InvalidAgentResponseError si the domain representation of this error *and* contains the error code for the JSON-RPC, it could also contains the status code for REST.

PR #503, opened 2025-12-03
fix: Remove unnecessary MapStruct conversion

kabir:


kabir:
Closing since Gemini is right, and the workarounds are ugly

PR #514, opened 2025-12-08
refactor: Complete migration from Jackson to Gson across codebase

### LINKED ISSUE #487: [Feat]: Remove Jackson annotation and use protobuf as the serialization mechanism
The https://github.com/a2aproject/A2A project now uses the a2a.proto file as the main source of truth.

Since we are already generating classes from this proto file, and using those for grpc and rest, it makes sense to use this for jsonrpc too.

The generated classes are very verbose and thus not suitable for exposing to users, so we will keep the domain model classes in the spec/ module (although perhaps this should have a new name now) as the user API, and perform mapping between these classes and the generated ones.

Internally we will use our [`ProtoUtils`](https://github.com/a2aproject/a2a-java/blob/main/spec-grpc/src/main/java/io/a2a/grpc/utils/ProtoUtils.java) class to convert between our domain model classes and the protobuf generated classes.

This means we no longer need to rely on Jackson for serializing the JSON files, so we can drop dependency on that, and remove the Jackson annotations from the domain model classes.

‼️  **As part of this, we should consider documenting the use of ProtoUtils etc. to help other integrators be able to serialize.**

ehsavoie:
Replace Jackson ObjectMapper with Gson throughout the codebase to eliminate Jackson dependency and provide custom serialization control.
    
Key Changes:
    - Created JsonUtil.java with Gson-based OBJECT_MAPPER singleton
    - Implemented 9 custom TypeAdapters for complex type handling:
        * OffsetDateTimeTypeAdapter - ISO-8601 datetime serialization
        * JSONRPCErrorTypeAdapter - Polymorphic error deserialization (12 error types)
        * ThrowableTypeAdapter - Avoids Java 17+ reflection restrictions
        * TaskStateTypeAdapter - Wire format enum serialization
        * RoleTypeAdapter - Message.Role enum handling
        * PartKindTypeAdapter - Part.Kind enum handling
        * PartTypeAdapter - Polymorphic Part deserialization (Text/File/Data)
        * StreamingEventKindTypeAdapter - Event type deserialization
        * FileContentTypeAdapter - FileWithBytes/FileWithUri discrimination
    - Added TaskSerializationTest with 21 comprehensive tests covering:
        * Round-trip serialization for all Task components
        * Direct JSON deserialization tests
        * All Part types and FileContent variants
    - Removed Jackson-specific deserializers and annotations
    - Updated all usages across 142 files (client, server, extras, tests)
    - Removed Jackson dependencies from spec/pom.xml
    
Technical Details:
    - Thread-safe singleton OBJECT_MAPPER with all adapters registered
    - Handles polymorphic types via two-pass parsing strategy
    - Maintains wire format compatibility with existing JSON-RPC protocol
    - Preserves null-safety with NullAway/JSpecify conventions

Fixes #487 🦕


spec-grpc/src/main/java/io/a2a/grpc/utils/JSONRPCUtils.java
```diff
@@ -509,7 +508,7 @@ public static String toJsonRPCResultResponse(Object requestId, com.google.protob
                 if (requestId instanceof String string) {
                     output.name("id").value(string);
                 } else if (requestId instanceof Number number) {
-                    output.name("id").value(number);
+                    output.name("id").value(Integer.toString(number.intValue()));
```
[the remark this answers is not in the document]
ehsavoie: With the suggested change the integer is serialized as 1.0 which is deserialized as a double and not an int

spec-grpc/src/main/java/io/a2a/grpc/utils/JSONRPCUtils.java
```diff
@@ -531,7 +530,7 @@ public static String toJsonRPCErrorResponse(Object requestId, JSONRPCError error
...
```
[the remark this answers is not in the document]
ehsavoie: With the suggested change the integer is serialized as 1.0 which is deserialized as a double and not an int

extras/task-store-database-jpa/src/test/java/io/a2a/extras/taskstore/database/jpa/JpaDatabaseTaskStoreTest.java
```diff
@@ -187,7 +187,7 @@ public void testTaskWithComplexMetadata() {
         assertEquals("test-task-5", retrieved.getId());
         assertNotNull(retrieved.getMetadata());
         assertEquals("value1", retrieved.getMetadata().get("key1"));
-        assertEquals(42, retrieved.getMetadata().get("key2"));
+        assertEquals(42, ((Number)retrieved.getMetadata().get("key2")).intValue());
```
kabir, line 190: This makes it look like it serialises `{"key2": "42"}` instead of `{"key2"; 42}`. Which might be fine, but also a bit strange

reference/jsonrpc/src/main/java/io/a2a/server/apps/quarkus/A2AServerRoutes.java
```diff
@@ -115,9 +106,18 @@ public void invokeJSONRPCHandler(@Body String body, RoutingContext rc) {
             }
         } catch (JSONRPCError e) {
             error = new JSONRPCErrorResponse(e);
+        } catch (io.a2a.spec.InvalidParamsJsonMappingException e) {
```
kabir: Use import rather than FQN

server-common/src/main/java/io/a2a/server/tasks/BasePushNotificationSender.java
```diff
@@ -80,10 +81,7 @@ private boolean dispatchNotification(Task task, PushNotificationConfig pushInfo)
...
         String body;
         try {
-            body = Utils.OBJECT_MAPPER.writeValueAsString(task);
-        } catch (JsonProcessingException e) {
-            LOGGER.debug("Error writing value as string: {}", e.getMessage(), e);
-            return false;
+            body = JsonUtil.toJson(task);
         } catch (Throwable throwable) {
             LOGGER.debug("Error writing value as string: {}", throwable.getMessage(), throwable);
             return false;
```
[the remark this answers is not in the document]
kabir: The end result is the same

spec-grpc/src/main/java/io/a2a/grpc/utils/JSONRPCUtils.java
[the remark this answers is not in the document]
kabir: Does something like https://www.baeldung.com/java-gson-prevent-expressing-integers-as-floats help at all?

spec/src/main/java/io/a2a/spec/A2AClientJSONError.java
```diff
@@ -17,7 +17,7 @@
  * <pre>{@code
  * try {
  *     AgentCard card = objectMapper.readValue(json, AgentCard.class);
- * } catch (JsonProcessingException e) {
+ * } catch (io.a2a.json.JsonProcessingException e) {
```
kabir, line 20: Can use an import

ehsavoie: That's a comment ;)

spec/pom.xml
```diff
@@ -21,14 +21,10 @@
...
         </dependency>
-
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-        </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.datatype</groupId>
-            <artifactId>jackson-datatype-jsr310</artifactId>
+            <groupId>com.google.code.gson</groupId>
```
jmesnil, line 25: the Jackson dependencies should also be removed from the parent pom's `<dependencyManagement>`

server-common/src/test/java/io/a2a/server/requesthandlers/AbstractA2ARequestHandlerTest.java
```diff
@@ -199,8 +200,11 @@ public PostBuilder body(String body) {
 
             @Override
             public A2AHttpResponse post() throws IOException, InterruptedException {
-                tasks.add(Utils.OBJECT_MAPPER.readValue(body, Task.TYPE_REFERENCE));
                 try {
+                    System.out.println("TestPostBuilder.post() body: " + body);
```
jmesnil: to remove

server-common/src/test/java/io/a2a/server/requesthandlers/AbstractA2ARequestHandlerTest.java
```diff
@@ -217,8 +221,14 @@ public String body() {
                             return "";
                         }
                     };
+                } catch (JsonProcessingException e) {
+                    System.err.println("Failed to parse task JSON: " + e.getMessage());
```
jmesnil: idem

server-common/src/test/java/io/a2a/server/requesthandlers/AbstractA2ARequestHandlerTest.java
```diff
@@ -199,8 +200,11 @@ public PostBuilder body(String body) {
 
             @Override
             public A2AHttpResponse post() throws IOException, InterruptedException {
-                tasks.add(Utils.OBJECT_MAPPER.readValue(body, Task.TYPE_REFERENCE));
                 try {
+                    System.out.println("TestPostBuilder.post() body: " + body);
+                    Task task = JsonUtil.fromJson(body, Task.class);
+                    tasks.add(task);
+                    System.out.println("Successfully parsed task: " + task.getId());
```
jmesnil: idem

spec-grpc/src/main/java/io/a2a/grpc/utils/JSONRPCUtils.java
```diff
@@ -509,7 +511,7 @@ public static String toJsonRPCResultResponse(Object requestId, com.google.protob
                 if (requestId instanceof String string) {
                     output.name("id").value(string);
                 } else if (requestId instanceof Number number) {
-                    output.name("id").value(number);
+                    output.name("id").value(number.intValue());
```
jmesnil: I'm not sure that an `int` is correct here? What if the id is bigger than `Integer.MAX`?
Should we use instead a long?

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+     * <p>
+     * This mapper is configured with strict parsing mode and all necessary custom TypeAdapters
+     * for A2A Protocol types including polymorphic types, enums, and date/time types.
+     * <p>
+     * Used throughout the SDK for consistent JSON serialization and deserialization.
+     *
+     * @see GsonFactory#createGson()
+     */
+    public static final Gson OBJECT_MAPPER = new GsonBuilder()
+            .setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)
```
jmesnil: related to my previous comment, if we map number to long or double, we should read id as long rather than int?

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+     *
+     * @see TaskState
+     * @see TaskState#asString()
+     * @see TaskState#fromString(String)
+     */
+    static class TaskStateTypeAdapter extends TypeAdapter<TaskState> {
+
+        @Override
+        public void write(JsonWriter out, TaskState value) throws java.io.IOException {
+            if (value == null) {
```
jmesnil, line 430: when would a TaskState be `null`? 
If there is such a case, shouldn't be serialised as `UNKNOWN` instead?

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+            } else {
+                out.value(value.asString());
+            }
+        }
+
+        @Override
+        public @Nullable TaskState read(JsonReader in) throws java.io.IOException {
+            if (in.peek() == com.google.gson.stream.JsonToken.NULL) {
+                in.nextNull();
+                return null;
```
jmesnil, line 442: idem

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+     * For deserialization, it parses the string to the enum constant.
+     *
+     * @see Message.Role
+     * @see Message.Role#asString()
+     */
+    static class RoleTypeAdapter extends TypeAdapter<Message.Role> {
+
+        @Override
+        public void write(JsonWriter out, Message.Role value) throws java.io.IOException {
+            if (value == null) {
```
jmesnil, line 470: idem

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+            } else {
+                out.value(value.asString());
+            }
+        }
+
+        @Override
+        public Message.@Nullable Role read(JsonReader in) throws java.io.IOException {
+            if (in.peek() == com.google.gson.stream.JsonToken.NULL) {
+                in.nextNull();
+                return null;
```
jmesnil, line 481: idem

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+            }
+
+            // Read the JSON as a tree so we can inspect the "kind" field
+            com.google.gson.JsonElement jsonElement = com.google.gson.JsonParser.parseReader(in);
+            if (!jsonElement.isJsonObject()) {
+                throw new JsonSyntaxException("Part must be a JSON object");
+            }
+
+            com.google.gson.JsonObject jsonObject = jsonElement.getAsJsonObject();
+            com.google.gson.JsonElement kindElement = jsonObject.get("kind");
```
jmesnil, line 599: When do we have a `kind` field in JSON? 
That's no longer part of the 1.0 spec afaict

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+            if (kindElement == null || !kindElement.isJsonPrimitive()) {
+                throw new JsonSyntaxException("Part must have a 'kind' field");
```
jmesnil, line 601: that's no longer true in 1.0 -> https://a2a-protocol.org/latest/specification/#a21-breaking-change-kind-discriminator-removed

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+
+            com.google.gson.JsonObject jsonObject = jsonElement.getAsJsonObject();
+            com.google.gson.JsonElement kindElement = jsonObject.get("kind");
+            if (kindElement == null || !kindElement.isJsonPrimitive()) {
+                throw new JsonSyntaxException("StreamingEventKind must have a 'kind' field");
+            }
+
+            String kind = kindElement.getAsString();
+            // Use the delegate Gson to deserialize to the concrete type
+            return switch (kind) {
```
jmesnil, line 683: code does not reflect the current 1.0 spec that no longer has a `kind`
and the field names are in camelCase now: https://a2a-protocol.org/latest/specification/#433-push-notification-payload

spec/src/main/java/io/a2a/json/JsonUtil.java
```diff
@@ -0,0 +1,729 @@
...
+            // Read the JSON as a tree to inspect the fields
+            com.google.gson.JsonElement jsonElement = com.google.gson.JsonParser.parseReader(in);
+            if (!jsonElement.isJsonObject()) {
+                throw new JsonSyntaxException("FileContent must be a JSON object");
+            }
+
+            com.google.gson.JsonObject jsonObject = jsonElement.getAsJsonObject();
+
+            // Distinguish between FileWithBytes and FileWithUri by checking for "bytes" or "uri" field
+            if (jsonObject.has("bytes")) {
```
jmesnil, line 753: is that correct?
The FilePart has fileWithBytes and fileWithUri fields -> https://a2a-protocol.org/latest/specification/#417-filepart

spec/src/main/java/io/a2a/spec/AgentCardSignature.java
```diff
@@ -34,9 +30,7 @@
  * @see <a href="https://tools.ietf.org/html/rfc7515">RFC 7515 - JSON Web Signature</a>
  * @see <a href="https://a2a-protocol.org/latest/">A2A Protocol Specification</a>
  */
-@JsonInclude(JsonInclude.Include.NON_ABSENT)
-@JsonIgnoreProperties(ignoreUnknown = true)
-public record AgentCardSignature(Map<String, Object> header, @JsonProperty("protected") String protectedHeader,
+public record AgentCardSignature(Map<String, Object> header, @com.google.gson.annotations.SerializedName("protected")String protectedHeader,
```
jmesnil: can you import or is there a reason to use FQN here?

spec/src/main/java/io/a2a/spec/DataPart.java
jmesnil, line 43: Kind is no longer used by the spec, do we need to keep a field for it or revisit the structure of our payloads?

spec/src/main/java/io/a2a/spec/HTTPAuthSecurityScheme.java
```diff
@@ -55,12 +46,10 @@ public HTTPAuthSecurityScheme(String bearerFormat, String scheme, String descrip
         this(bearerFormat, scheme, description, HTTP);
     }
 
-    @JsonCreator
-    public HTTPAuthSecurityScheme(@JsonProperty("bearerFormat") String bearerFormat, @JsonProperty("scheme") String scheme,
-                                  @JsonProperty("description") String description, @JsonProperty("type") String type) {
+    public HTTPAuthSecurityScheme(String bearerFormat, String scheme, String description, String type) {
```
jmesnil, line 49: nitpick but I don't get why we need to pass a `type` here as the type of the scheme is defined by the class.

spec/src/main/java/io/a2a/spec/ListTasksRequest.java
```diff
@@ -4,25 +4,16 @@
...
-@JsonIgnoreProperties(ignoreUnknown = true)
 public final class ListTasksRequest extends NonStreamingJSONRPCRequest<ListTasksParams> {
 
     public static final String METHOD = "ListTask";
 
-    @JsonCreator
-    public ListTasksRequest(@JsonProperty("jsonrpc") String jsonrpc, @JsonProperty("id") Object id,
-                            @JsonProperty("method") String method, @JsonProperty("params") ListTasksParams params) {
+    public ListTasksRequest(String jsonrpc, Object id, String method, ListTasksParams params) {
         if (jsonrpc == null || jsonrpc.isEmpty()) {
```
jmesnil, line 17: nitpick, the validation differs from https://github.com/a2aproject/a2a-java/pull/514/files#diff-1a074a4d32c70956ec3ec7aae747489990148b90c64d043c0e35f21c2e1b8ef5R26 that checks the version value too

spec/src/main/java/io/a2a/spec/Message.java
```diff
@@ -57,12 +45,11 @@ public Message(Role role, List<Part<?>> parts, String messageId, String contextI
...
-    public Message(@JsonProperty("role") Role role, @JsonProperty("parts") List<Part<?>> parts,
-                   @JsonProperty("messageId") String messageId, @JsonProperty("contextId") String contextId,
-                   @JsonProperty("taskId") String taskId, @JsonProperty("referenceTaskIds") List<String> referenceTaskIds,
-                   @JsonProperty("metadata") Map<String, Object> metadata, @JsonProperty("extensions") List<String> extensions,
-                   @JsonProperty("kind") String kind) {
+    public Message(Role role, List<Part<?>> parts,
+                   String messageId, String contextId,
+                   String taskId, List<String> referenceTaskIds,
+                   Map<String, Object> metadata, List<String> extensions,
+                   String kind) {
```
jmesnil, line 52: again, I question the use of `kind` here

spec/src/main/java/io/a2a/spec/Message.java
jmesnil, line 138: According to the spec, role can also be `ROLE_UNSPECIFIED` https://a2a-protocol.org/latest/specification/#415-role

Should we also use the same values that the spec for this domain enum (eg `ROLE_USER`)

spec/src/main/java/io/a2a/spec/MutualTLSSecurityScheme.java
```diff
@@ -49,9 +41,7 @@ public MutualTLSSecurityScheme() {
         this(null, MUTUAL_TLS);
     }
 
-    @JsonCreator
-    public MutualTLSSecurityScheme(@JsonProperty("description") String description,
-                                   @JsonProperty("type") String type) {
+    public MutualTLSSecurityScheme(String description, String type) {
```
jmesnil, line 44: `type` should not be needed

spec/src/main/java/io/a2a/spec/OAuth2SecurityScheme.java
```diff
@@ -39,9 +30,7 @@ public OAuth2SecurityScheme(OAuthFlows flows, String description, String oauth2M
         this(flows, description, oauth2MetadataUrl, OAUTH2);
     }
 
-    @JsonCreator
-    public OAuth2SecurityScheme(@JsonProperty("flows") OAuthFlows flows, @JsonProperty("description") String description,
-                                @JsonProperty("oauth2MetadataUrl") String oauth2MetadataUrl, @JsonProperty("type") String type) {
+    public OAuth2SecurityScheme(OAuthFlows flows, String description, String oauth2MetadataUrl, String type) {
```
jmesnil, line 33: idem

spec/src/main/java/io/a2a/spec/OpenIdConnectSecurityScheme.java
```diff
@@ -47,9 +39,7 @@ public OpenIdConnectSecurityScheme(String openIdConnectUrl, String description)
         this(openIdConnectUrl, description, OPENID_CONNECT);
     }
 
-    @JsonCreator
-    public OpenIdConnectSecurityScheme(@JsonProperty("openIdConnectUrl") String openIdConnectUrl,
-                                       @JsonProperty("description") String description, @JsonProperty("type") String type) {
+    public OpenIdConnectSecurityScheme(String openIdConnectUrl, String description, String type) {
```
jmesnil, line 42: idem

spec/src/main/java/io/a2a/spec/OpenIdConnectSecurityScheme.java
```diff
@@ -33,9 +28,6 @@
  * @see <a href="https://openid.net/specs/openid-connect-discovery-1_0.html">OpenID Connect Discovery</a>
  * @see <a href="https://a2a-protocol.org/latest/">A2A Protocol Specification</a>
  */
-@JsonTypeName(OPENID_CONNECT)
-@JsonInclude(JsonInclude.Include.NON_ABSENT)
-@JsonIgnoreProperties(ignoreUnknown = true)
 public final class OpenIdConnectSecurityScheme implements SecurityScheme {
 
     public static final String OPENID_CONNECT = "openIdConnect";
```
jmesnil, line 33: should that be  `openIdConnectSecurityScheme` according to https://a2a-protocol.org/latest/specification/#451-securityscheme

spec/src/main/java/io/a2a/spec/SendMessageRequest.java
```diff
@@ -56,9 +49,7 @@ public final class SendMessageRequest extends NonStreamingJSONRPCRequest<Message
      * @param params the message send parameters (required)
      * @throws IllegalArgumentException if validation fails
      */
-    @JsonCreator
-    public SendMessageRequest(@JsonProperty("jsonrpc") String jsonrpc, @JsonProperty("id") Object id,
-                              @JsonProperty("method") String method, @JsonProperty("params") MessageSendParams params) {
+    public SendMessageRequest(String jsonrpc, Object id, String method, MessageSendParams params) {
```
jmesnil, line 52: nitpick, we should have a helper method to make sure we validate the `jsonrpc` field consistently across all the request classes

spec/src/main/java/io/a2a/spec/StreamingEventKind.java
jmesnil, line 44: this should no longer be needed

spec/src/main/java/io/a2a/spec/TaskNotCancelableError.java
```diff
@@ -35,26 +30,23 @@
...
-@JsonInclude(JsonInclude.Include.NON_ABSENT)
-@JsonIgnoreProperties(ignoreUnknown = true)
 public class TaskNotCancelableError extends JSONRPCError {
 
     public TaskNotCancelableError() {
         this(null, null, null);
     }
 
-    @JsonCreator
     public TaskNotCancelableError(
```
jmesnil, line 39: there should not be a need for a `code` parameter as the type of error is represented by the class itself.

spec/src/main/java/io/a2a/spec/TaskState.java
```diff
@@ -104,14 +100,13 @@ public boolean isFinal(){
...
-     * This method is used by Jackson to deserialize TaskState values from their
+     * This method is used to deserialize TaskState values from their
      * wire format during JSON parsing.
      *
      * @param state the string representation of the state
      * @return the corresponding TaskState enum constant
      * @throws IllegalArgumentException if the state string is not recognized
      */
-    @JsonCreator
     public static TaskState fromString(String state) {
```
jmesnil, line 110: should we use the same values for this domain class than the values in the spec -> https://a2a-protocol.org/latest/specification/#413-taskstate

spec/src/main/java/io/a2a/spec/TaskStatusUpdateEvent.java
```diff
@@ -15,15 +11,13 @@
  * An event sent by the agent to notify the client of a change in a task's status.
  * This is typically used in streaming or subscription models.
  */
-@JsonTypeName(STATUS_UPDATE)
-@JsonInclude(JsonInclude.Include.NON_ABSENT)
-@JsonIgnoreProperties(ignoreUnknown = true)
 public final class TaskStatusUpdateEvent implements EventKind, StreamingEventKind, UpdateEvent {
 
     public static final String STATUS_UPDATE = "status-update";
```
jmesnil, line 16: idem (use the same value than in the spec)

spec/src/main/java/io/a2a/util/Utils.java
```diff
@@ -25,50 +26,33 @@
...
      * @param <T> the target type
      * @param data JSON string to deserialize
-     * @param typeRef type reference specifying the target type (for generic types)
+     * @param typeRef class reference specifying the target type
      * @return deserialized object of type T
      * @throws JsonProcessingException if JSON parsing fails
      */
-    public static <T> T unmarshalFrom(String data, TypeReference<T> typeRef) throws JsonProcessingException {
-        return OBJECT_MAPPER.readValue(data, typeRef);
+    public static <T> T unmarshalFrom(String data, Class<T> typeRef) throws JsonProcessingException {
```
jmesnil, line 54: do we need these helper methods now that most of the code is using `JsonUtil`?

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
```diff
@@ -0,0 +1,713 @@
...
+                .status(new TaskStatus(TaskState.COMPLETED))
+                .artifacts(List.of(artifact))
+                .build();
+
+        // Serialize
+        String json = JsonUtil.toJson(task);
+
+        // Verify JSON contains artifact data
+        assertTrue(json.contains("\"artifactId\":\"artifact-1\""));
+        assertTrue(json.contains("Hello"));
```
jmesnil, line 92: that might not be enough. Does the json contains `{"text": "Hello"}`?

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
```diff
@@ -0,0 +1,713 @@
...
+                .status(new TaskStatus(TaskState.WORKING))
+                .history(List.of(message))
+                .build();
+
+        // Serialize
+        String json = JsonUtil.toJson(task);
+
+        // Verify JSON contains history data
+        assertTrue(json.contains("\"role\":\"user\""));
+        assertTrue(json.contains("Test message"));
```
jmesnil, line 124: idem

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
```diff
@@ -0,0 +1,713 @@
...
+                    .id("task-" + state.asString())
+                    .contextId("context-123")
+                    .status(new TaskStatus(state))
+                    .build();
+
+            // Serialize
+            String json = JsonUtil.toJson(task);
+
+            // Verify state is serialized correctly
+            assertTrue(json.contains("\"state\":\"" + state.asString() + "\""));
```
jmesnil, line 193: that's not correct, I would expect to get a value that is compliant with the spec

`"state": "TASK_STATE_SUBMITTED"`

and not

`"state": "submitted"`

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
```diff
@@ -0,0 +1,713 @@
...
+                .contextId("context-456")
+                .status(new TaskStatus(TaskState.COMPLETED))
+                .artifacts(List.of(artifact))
+                .build();
+
+        // Serialize
+        String json = JsonUtil.toJson(task);
+
+        // Verify JSON contains file part data
+        assertTrue(json.contains("\"kind\":\"file\""));
```
jmesnil, line 250: not spec compliant

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
```diff
@@ -0,0 +1,713 @@
...
+        Task deserialized = JsonUtil.fromJson(json, Task.class);
+
+        // Verify file part is preserved
+        Part<?> part = deserialized.getArtifacts().get(0).parts().get(0);
+        assertTrue(part instanceof FilePart);
+        FilePart deserializedFilePart = (FilePart) part;
+        assertTrue(deserializedFilePart.getFile() instanceof FileWithBytes);
+        FileWithBytes fileWithBytes = (FileWithBytes) deserializedFilePart.getFile();
+        assertEquals("document.pdf", fileWithBytes.name());
+        assertEquals("application/pdf", fileWithBytes.mimeType());
```
jmesnil, line 264: btw, `mimeType` should be renamed to `mediaType` (https://a2a-protocol.org/latest/specification/#417-filepart)

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
```diff
@@ -0,0 +1,713 @@
...
+                .contextId("context-456")
+                .status(new TaskStatus(TaskState.COMPLETED))
+                .artifacts(List.of(artifact))
+                .build();
+
+        // Serialize
+        String json = JsonUtil.toJson(task);
+
+        // Verify JSON contains data part
+        assertTrue(json.contains("\"kind\":\"data\""));
```
jmesnil, line 321: not compliant

jmesnil (changes requested):
The removal of Jackson sounds good (it just needs to be removed from the parent pom too) but I'm not sure that the serialisation is correct (eg the `kind` field)

In the tests, we verify that we can unserialize/deserialize the object but the JSON strings does not comply with the 1.0 spec

More generally, there are too many unnecessary fields that complicates the API (`code` in Error, `kind` in payload, `type` in other classes

I also wonder if the domain class that have String values (`Role`, `TaskState`) should be updated to use the same values that the spec.

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
ehsavoie: This is a domain object serialization, not a protocol payload

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
ehsavoie: This is a domain object serialization, not a protocol payload

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
ehsavoie: This is a domain object serialization, not a protocol payload

jmesnil:
another note on my review, I've been working on updating the TCK after #484 and a lot of tests passes so we are able to receive/return valid JSON from our transports.

My comments about producing invalid JSON payload that is not compliant to the spec is not correct.
AIUI,  this json serialisation is "internal" to the project and only serialised through the wire for replications and not for end-to-end communication.
Still, I wonder if we should have a internal JSON representation that is consistent with the spec.

spec-grpc/src/main/java/io/a2a/grpc/utils/JSONRPCUtils.java
ehsavoie: yes, should be better but the ID is supposed to be a String or an Integer

jmesnil: is it?

According to https://www.jsonrpc.org/specification#request_object:

> An identifier established by the Client that MUST contain a String, Number, or NULL value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [[1]](https://www.jsonrpc.org/specification#id1) and Numbers SHOULD NOT contain fractional parts [[2]](https://www.jsonrpc.org/specification#id2)

and a JSON Number is not tied to a Java `int` (https://www.json.org/json-en.html)

jmesnil:
one thing that trips me up is that all my comments are for the "internal" JSON payload that is not used by clients or server but is exposed by the API through the io.a2a.util.Utils class.
If that JSON representation is internal, it should not leak to the API.

spec/src/test/java/io/a2a/spec/TaskSerializationTest.java
jmesnil: and I'm wrong anyway, the json serialisation must use `submitted` according to https://a2a-protocol.org/latest/specification/#55-json-field-naming-convention

spec/src/main/java/io/a2a/spec/HTTPAuthSecurityScheme.java
ehsavoie: Maybe that was for Jackson

ehsavoie: It seems required for gson (de)serialization as well

spec/src/main/java/io/a2a/spec/ListTasksRequest.java
jmesnil: tracked by #522

spec/src/main/java/io/a2a/spec/HTTPAuthSecurityScheme.java
jmesnil: The ctor parameter is required or the `type` field?

spec/src/main/java/io/a2a/spec/SendMessageRequest.java
jmesnil: tracked by #522

jmesnil:
@ehsavoie This review raised a few questions I have over the design of the code and I created #523 to clarify them.

jmesnil:
@ehsavoie thanks!

PR #525, opened 2025-12-10
chore: Enable javadoc in the default profile

### LINKED ISSUE #524: chore: Update github action on PR to detect javadoc errors
Currently the job build with "-Prelease" is on pull_request_target instead of pull_request so we have the error too late

jmesnil:
Any javadoc errors (not warnings) will make
Maven fail

This fixes #524

### DIFF
--- a/spec/src/main/java/io/a2a/util/Utils.java
+++ b/spec/src/main/java/io/a2a/util/Utils.java
@@ -44,7 +44,7 @@ public class Utils {
     /**
      * Deserializes JSON string into a typed object using Gson.
      * <p>
-     * This method uses the pre-configured {@link JsonUtil#fromJson(String, Type)} to parse JSON.
+     * This method uses the pre-configured {@link JsonUtil#fromJson(String, Class)} to parse JSON.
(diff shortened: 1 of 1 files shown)


kabir:
@jmesnil I disagree with this fix. We should fix the GH Action instead.

jmesnil:
@kabir Do you disagree with generating Javadoc when the project is built (with the default profile)?

The GitHub action runs on `pull_request_target` to safely access the project secrets.
If we want to verify the Javadoc on PR, we need to have a GH action that targets the pull request. We could add another GH action with a dedicated `javadoc` profile (which would be disabled from the default profile) or run the `Build and Test` job with that `-Pjavadoc` profile

jmesnil:
@kabir I've opened #529 to remediate my incorrect PR. Could you review if that alleviates your concern?

kabir:
@jmesnil Yeah, my main concern is about running this locally

a) it is probably takes some time 
b) all the warnings are noisy

I'll comment on #529 for the rest

Issue #522, opened 2025-12-10
[Bug]: Consolidate validation of JSON-RPC requests


jmesnil:
### What happened?

In the `spec` module, the `XXXRequest` have different validation for the JSON-RPC methods:

```java
        if (jsonrpc == null || jsonrpc.isEmpty()) {
            throw new IllegalArgumentException("JSON-RPC protocol version cannot be null or empty");
        }
        if (jsonrpc != null && !jsonrpc.equals(JSONRPC_VERSION)) {
            throw new IllegalArgumentException("Invalid JSON-RPC protocol version");
        }
        Assert.checkNotNullParam("method", method);
        if (!method.equals(METHOD)) {
            throw new IllegalArgumentException("Invalid ListTasksRequest method");
        }
        Assert.checkNotNullParam("params", params);
        Assert.isNullOrStringOrInteger(id);
```

```java
        if (jsonrpc != null && ! jsonrpc.equals(JSONRPC_VERSION)) {
            throw new IllegalArgumentException("Invalid JSON-RPC protocol version");
        }
        Assert.checkNotNullParam("method", method);
        if (! method.equals(METHOD)) {
            throw new IllegalArgumentException("Invalid GetAuthenticatedExtendedCardRequest method");
        }
        Assert.isNullOrStringOrInteger(id);
        this.jsonrpc = Utils.defaultIfNull(jsonrpc, JSONRPC_VERSION);
```

All request classes should use the same helper method to standardise the validation of the JSON-RPC spec in a consistent way

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #527, opened 2025-12-10
docs: Add comprehensive Javadoc to spec module classes

ehsavoie:
Added Javadoc documentation for:
- Public constants (error codes, method names, type identifiers)
- Generic type parameters in base request/response classes
- Record compact constructors (12 record classes)
- Exception constructors (5 exception classes)
- APIKeySecurityScheme (enum, methods, constructors, Builder)
- Builder classes and methods (CancelTaskRequest, DeleteTaskPushNotificationConfig*)
- Request/response constructors

This resolves all Javadoc warnings and errors  in the spec module, enabling clean builds with the -Prelease profile.

### DIFF
--- a/spec/src/main/java/io/a2a/spec/AgentCardSignature.java
+++ b/spec/src/main/java/io/a2a/spec/AgentCardSignature.java
@@ -34,6 +34,11 @@ import io.a2a.util.Assert;
 public record AgentCardSignature(Map<String, Object> header, @SerializedName("protected")String protectedHeader,
                                  String signature) {
 
+    /**
+     * Compact constructor that validates required fields.
+     *
+     * @throws IllegalArgumentException if protectedHeader or signature is null
+     */
     public AgentCardSignature {
         Assert.checkNotNullParam("protectedHeader", protectedHeader);
         Assert.checkNotNullParam("signature", signature);
@@ -55,6 +60,12 @@ public record AgentCardSignature(Map<String, Object> header, @SerializedName("pr
         private Map<String, Object> header;
         String protectedHeader;
         String signature;
+
+        /**
+         * Creates a new Builder with all fields unset.
+         */
+        public Builder() {
+        }
 
         /**
          * Sets the optional unprotected header with additional metadata.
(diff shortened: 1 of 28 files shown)


spec/src/main/java/io/a2a/spec/AgentCardSignature.java
```diff
@@ -34,6 +34,11 @@
 public record AgentCardSignature(Map<String, Object> header, @SerializedName("protected")String protectedHeader,
                                  String signature) {
 
+    /**
+     * Compact constructor that validates required fields.
+     *
+     * @throws IllegalArgumentException if protectedHeader or signature is null
```
jmesnil, line 40: could we add @NonNull to this record fields instead?

spec/src/main/java/io/a2a/spec/CancelTaskRequest.java
```diff
@@ -26,8 +26,18 @@
  */
 public final class CancelTaskRequest extends NonStreamingJSONRPCRequest<TaskIdParams> {
 
+    /** The JSON-RPC method name for canceling tasks. */
     public static final String METHOD = "CancelTask";
 
+    /**
+     * Creates a new CancelTaskRequest with the specified JSON-RPC parameters.
+     *
+     * @param jsonrpc the JSON-RPC version (defaults to "2.0" if null)
```
jmesnil, line 35: that's not correct, if `null`, the exception will be raised (we should fix that with #522)

spec/src/main/java/io/a2a/spec/AgentCardSignature.java
ehsavoie: I would do that in a different step

spec/src/main/java/io/a2a/spec/CancelTaskRequest.java
ehsavoie: Hum I disagree, which exception should be raised here ?
if (jsonrpc != null && ! JSONRPC_VERSION.equals(jsonrpc)) {
            throw new IllegalArgumentException("Invalid JSON-RPC protocol version");
        }
will not be thrown and this.jsonrpc = defaultIfNull(jsonrpc, JSONRPC_VERSION);

Issue #557, opened 2025-12-19
Intermittent failure in QuarkusA2AJSONRPCTest/QuarkusA2AGrpcTest/QuarkusA2ARestTest.testNonBlockingWithMultipleMessages()


kabir:
e.g. here https://github.com/a2aproject/a2a-java/actions/runs/20375459905/job/58552633503?pr=556#step:4:3842

These all extend AbstractA2AServerTest, which contains the code, so it is transport independent. Probably a race condition in the test

PR #566, opened 2026-01-02
fix: Event consumer should stop on input_required

cmtoan:
# Description

Event consumer should stop on input_required task as fixed in the python version https://github.com/a2aproject/a2a-python/pull/167

### DIFF
--- a/server-common/src/main/java/io/a2a/server/events/EventConsumer.java
+++ b/server-common/src/main/java/io/a2a/server/events/EventConsumer.java
@@ -95,7 +95,7 @@ public class EventConsumer {
                         }
 
                         if (isFinalEvent) {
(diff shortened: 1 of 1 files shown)


cmtoan:
There are flaky tests in QuarkusA2ARestTest.

ehsavoie:
@cmtoan May you add some javadoc on your isStreamTerminatingTask to explain why INPUT_REQUIRED is terminating.
maybe update the log message on line 98 as INPUT_REQUIRED is not a final event. Could you be more specific with the flaky tests and create an issue

cmtoan:
@ehsavoie yes, I add a javadoc and update the log.
For the flaky tests, I've seen there is already an issue here https://github.com/a2aproject/a2a-java/issues/557

PR #569, opened 2026-01-06
fix: reject gRPC message without a messageId

jmesnil:
Failure is reported in the TCK by the tests/optional/features/test_invalid_business_logic.py::test_missing_required_message_fields test.

### DIFF
--- a/spec-grpc/src/main/java/io/a2a/grpc/mapper/A2ACommonFieldMapper.java
+++ b/spec-grpc/src/main/java/io/a2a/grpc/mapper/A2ACommonFieldMapper.java
@@ -14,6 +14,8 @@ import com.google.protobuf.Value;
 import org.mapstruct.Mapper;
 import org.mapstruct.Named;
 
+import io.a2a.spec.InvalidParamsError;
+
 /**
  * Common field mapping utilities shared across all mappers.
  * <p>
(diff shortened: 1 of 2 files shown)


kabir:
I also saw this error, and was planning on working on it this morning. I think the TCK looks for an InvalidParamsError, so we need the mapper to throw that somehow

EDIT: nm, I see you have changed the mapper :-)

spec-grpc/src/main/java/io/a2a/grpc/mapper/A2ACommonFieldMapper.java
```diff
@@ -46,6 +46,24 @@ default String emptyToNull(String value) {
...
+     * Use this with {@code @Mapping(qualifiedByName = "requireNonEmpty")}.
+     *
+     * @param value the protobuf string value
+     * @return the value if not null/empty
+     * @throws IllegalArgumentException if value is null or empty
+     */
+    @Named("requireNonEmpty")
+    default String requireNonEmpty(String value) {
+        if (value == null || value.isEmpty()) {
+            throw new IllegalArgumentException("Required field cannot be null or empty");
```
kabir: I think this should be InvalidParamsError since that is what the TCK tests for

PR #570, opened 2026-01-06
fix: Fixing TCK support for extend card.

ehsavoie:
Some TCK tests were skipped because of that.

### DIFF
--- a/tck/src/main/java/io/a2a/tck/server/AgentCardProducer.java
+++ b/tck/src/main/java/io/a2a/tck/server/AgentCardProducer.java
@@ -2,6 +2,7 @@ package io.a2a.tck.server;
 
 import static io.a2a.spec.AgentCard.CURRENT_PROTOCOL_VERSION;
 
+import io.a2a.server.ExtendedAgentCard;
 import java.util.Collections;
 import java.util.List;
 
@@ -22,6 +23,7 @@ public class AgentCardProducer {
 
     @Produces
     @PublicAgentCard
+    @ExtendedAgentCard
     public AgentCard agentCard() {
 
         String sutJsonRpcUrl = getEnvOrDefault("SUT_JSONRPC_URL", DEFAULT_SUT_URL);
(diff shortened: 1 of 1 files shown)


tck/src/main/java/io/a2a/tck/server/AgentCardProducer.java
```diff
@@ -51,6 +53,7 @@ public AgentCard agentCard() {
                                 .examples(List.of("hi", "hello world"))
                                 .build()))
                 .protocolVersion(CURRENT_PROTOCOL_VERSION)
+                .supportsExtendedAgentCard(true)
```
jmesnil, line 56: I'm not sure to understand what's the @ExtendedAgentCard is doing?
What happens if the annotation is present and  `supportsExtendedAgentCard` is false?
Conversely, what happen if the annotation is not there but `supportsExtendedAgentCard` is true?

ehsavoie: Currently if the support is set to false then we don't need the annotation as the TCK will skip the extended card tests. If it is set to true without the annotation then it will fail the TCK because we don't provide an extended card and thus we are failing

ehsavoie: We should return a different card with a new method or say we don't support extended agent card

jmesnil: sorry but I still don't get it.
Why is there a need for an annotation? If `supportsExtendedAgentCard` is `false`, there is nothing to do. If it is `true`, then the SDK should create an handler for `GetExtendedAgentCard` (with authentication). 

What am I missing?

jmesnil: Answering my own question...

`ExtendedAgentCard` is the annotation to identify the extended agent card. If there is a producer for this annotation, it will be used by the transport to respond to the `getExtendedAgentCard` (provided that  supportsExtendedAgentCard  is true for the public agent card).
In the TCK, the same producer is used to provide the public agent card *and* the extended agent card from the same method.
In practice, this would be an anti pattern. I'm not sure if that makes sense either for the TCK but that should not block this PR....

PR #349, opened 2025-10-14
Async A2A Client

brasseld:
This PR is an idea of what can be an Async A2A client.

The main changes are about http-client and the transport layer which are fully asynchronous, based on JDK Future / CompletableFuture.

# Description

Thank you for opening a Pull Request!
Before submitting your PR, there are a few things you can do to make sure it goes smoothly:

- [ ] Follow the [`CONTRIBUTING` Guide](../CONTRIBUTING.md).
- [ ] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
    - Important Prefixes for [release-please](https://github.com/googleapis/release-please):
        - `fix:` which represents bug fixes, and correlates to a [SemVer](https://semver.org/) patch.
        - `feat:` represents a new feature, and correlates to a SemVer minor.
        - `feat!:`, or `fix!:`, `refactor!:`, etc., which represent a breaking change (indicated by the `!`) and will result in a SemVer major.
- [ ] Ensure the tests pass
- [ ] Appropriate READMEs were updated (if necessary)

Fixes #<issue_number_goes_here> 🦕

PR #584, opened 2026-01-13
feat: Add pluggable A2AHttpClient with Vert.x implementation

### LINKED ISSUE #583: [Feat]: Provide another other A2AHttpClient implementation

Currently only the JDK HTTP Client is supported for the A2AHttpClient implementation. We should be able to provide one using Vert.x which would make more sense in a Quarkus environment.

Use the Vert.x Web Client to provide such an implementation

_No response_

### Additional context

_No response_

### Code of Conduct

ehsavoie:
Introduce ServiceLoader-based A2AHttpClient client architecture with two implementations:
- JDK HttpClient (core, always available)
- Vert.x WebClient (optional, high-performance)

Changes:
- Create http-client-vertx module for Vert.x implementation
- Add ServiceLoader infrastructure (A2AHttpClientFactory, providers)
- Update dependent modules to use Vert.x client by default
- Add comprehensive tests and usage examples

Users can choose implementation via Maven dependencies. Priority-based selection: Vert.x (100) > JDK (0).

Fixes #583 🦕

### DIFF
--- a/client/base/pom.xml
+++ b/client/base/pom.xml
@@ -24,6 +24,11 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>a2a-java-sdk-http-client-vertx</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
             <artifactId>a2a-java-sdk-client-transport-spi</artifactId>
         </dependency>
         <dependency>
--- a/client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransportConfigBuilder.java
+++ b/client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransportConfigBuilder.java
@@ -1,7 +1,7 @@
 package io.a2a.client.transport.jsonrpc;
 
 import io.a2a.client.http.A2AHttpClient;
-import io.a2a.client.http.JdkA2AHttpClient;
+import io.a2a.client.http.A2AHttpClientFactory;
 import io.a2a.client.transport.spi.ClientTransportConfigBuilder;
 import org.jspecify.annotations.Nullable;
 
@@ -11,7 +11,7 @@ import org.jspecify.annotations.Nullable;
  * This builder provides a fluent API for configuring the JSON-RPC transport protocol.
  * All configuration options are optional - if not specified, sensible defaults are used:
  * <ul>
- *   <li><b>HTTP client:</b> {@link JdkA2AHttpClient} (JDK's built-in HTTP client)</li>
+ *   <li><b>HTTP client:</b> Auto-selected via {@link A2AHttpClientFactory} (prefers Vert.x, falls back to JDK)</li>
  *   <li><b>Interceptors:</b> None</li>
  * </ul>
  * <p>
@@ -78,7 +78,7 @@ public class JSONRPCTransportConfigBuilder extends ClientTransportConfigBuilder<
      *   <li>Custom header handling</li>
      * </ul>
      * <p>
-     * If not specified, the default {@link JdkA2AHttpClient} is used.
+     * If not specified, a client is auto-selected via {@link A2AHttpClientFactory}.
      * <p>
      * Example:
      * <pre>{@code
@@ -101,16 +101,16 @@ public class JSONRPCTransportConfigBuilder extends ClientTransportConfigBuilder<
     /**
      * Build the JSON-RPC transport configuration.
      * <p>
-     * If no HTTP client was configured, the default {@link JdkA2AHttpClient} is used.
+     * If no HTTP client was configured, one is auto-selected via {@link A2AHttpClientFactory}.
      * Any configured interceptors are transferred to the configuration.
      *
      * @return the configured JSON-RPC transport configuration
      */
     @Override
     public JSONRPCTransportConfig build() {
-        // No HTTP client provided, fallback to the default one (JDK-based implementation)
+        // No HTTP client provided, use factory to get best available implementation
         if (httpClient == null) {
-            httpClient = new JdkA2AHttpClient();
+            httpClient = A2AHttpClientFactory.create();
         }
 
         JSONRPCTransportConfig config = new JSONRPCTransportConfig(httpClient);
--- a/extras/http-client-vertx/src/main/java/io/a2a/client/http/VertxA2AHttpClient.java
+++ b/extras/http-client-vertx/src/main/java/io/a2a/client/http/VertxA2AHttpClient.java
@@ -1,0 +1,554 @@
+package io.a2a.client.http;
+
+import static java.net.HttpURLConnection.HTTP_FORBIDDEN;
+import static java.net.HttpURLConnection.HTTP_MULT_CHOICE;
+import static java.net.HttpURLConnection.HTTP_OK;
+import static java.net.HttpURLConnection.HTTP_UNAUTHORIZED;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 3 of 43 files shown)


ehsavoie:
@brasseld It has been quite some time but if you want to take a look at this PR :)

client/base/pom.xml
```diff
@@ -22,6 +22,10 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>a2a-java-sdk-http-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>a2a-java-sdk-http-client-vertx</artifactId>
```
jmesnil, line 27: if that module is always a dependency of the client, won't the Vert.x HTTP client *always* be used?

Shouldn't that dependency scope be `provided` and it's up to the client to include it to use the Vert.x HTTP client?

ehsavoie: Agreed and fixed

client/transport/jsonrpc/pom.xml
```diff
@@ -21,6 +21,10 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>a2a-java-sdk-http-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>a2a-java-sdk-http-client-vertx</artifactId>
```
jmesnil: idem, should be `provided`

pom.xml
```diff
@@ -511,6 +516,7 @@
         <module>extras/push-notification-config-store-database-jpa</module>
         <module>extras/queue-manager-replicated</module>
         <module>http-client</module>
+        <module>http-client-vertx</module>
```
jmesnil: as the use of the Vert.x is optional, should that module be moved to the `extras` directory?

brasseld:
Hey @ehsavoie , thanks for adding me here, it's been a while indeed.

Some general comments:

* Good to see this finally coming !
* I'm questioning the reason for relying on a Provider / ServiceLoader. I do feel it is limitting the developer to use the right client implementation based on the trasnport type. What if you want to leverage different client implementation based on the transports configured using the ClientBuilder (https://github.com/a2aproject/a2a-java/pull/347/changes#diff-2ab588dba6c8155a6c004136c62a56cb58c3ca0a37d7db909da1f933e4b3ff14R57) ? Also, (that's probably my background in API Gateways at Gravitee), but most of the time, it is very helpful to provide a way to fine tune the configuration of the HTTP client, so that you can configure pooling / proxy / security and all that sort of stuff when useful (https://github.com/a2aproject/a2a-java/pull/347/changes#diff-4f7ff5e67fcc2927b3ef362fd1adabdd49880696f9ed549190c3637c13a61136R10). 
* I'm not a huge fan of using Latch and AtomicReference with Vertx HTTP Client, but I guess this one is a much bigger discussion to have, about providing an async client implementation. Basically something I've been trying out a while back : https://github.com/a2aproject/a2a-java/pull/349

Good work in general, I will be giving it a try later today (US TZ).

ehsavoie:
@brasseld :
You can still provide the client directly, this SPI is the default fallback behaviour.
The provider enables the use of the vertx client with a fallback to the jdk client if vertx classes are not available.
Yeah, the issue was the connection could be closed (onComplete called) before the onError was called iirc

extras/http-client-vertx/src/main/java/io/a2a/client/http/VertxA2AHttpClient.java
```diff
@@ -0,0 +1,565 @@
...
+         * <ul>
+         * <li>Network errors (connection refused, timeout, etc.)</li>
+         * <li>HTTP 401 Unauthorized - with message from {@link A2AErrorMessages#AUTHENTICATION_FAILED}</li>
+         * <li>HTTP 403 Forbidden - with message from {@link A2AErrorMessages#AUTHORIZATION_FAILED}</li>
+         * </ul>
+         * @throws InterruptedException if the thread is interrupted while waiting
+         */
+        @Override
+        public A2AHttpResponse post() throws IOException, InterruptedException {
+            Buffer bodyBuffer = Buffer.buffer(body, StandardCharsets.UTF_8.name());
```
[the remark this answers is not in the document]
ehsavoie: You are wrong, this is not available on vertx 4.5

extras/http-client-vertx/src/main/java/io/a2a/client/http/VertxA2AHttpClient.java
```diff
@@ -0,0 +1,565 @@
...
+        }
+
+        @Override
+        public CompletableFuture<Void> postAsyncSSE(
+                Consumer<String> messageConsumer,
+                Consumer<Throwable> errorConsumer,
+                Runnable completeRunnable) throws IOException, InterruptedException {
+
+            HttpRequest<Buffer> request = webClient.postAbs(url);
+            Buffer bodyBuffer = Buffer.buffer(body, StandardCharsets.UTF_8.name());
```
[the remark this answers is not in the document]
ehsavoie: You are wrong, this is not available on vertx 4.5

jmesnil (approved):
LGTM

brasseld:
Sounds good.

Should the extras README be updated as well maybe ?  https://github.com/a2aproject/a2a-java/blob/main/extras/README.md

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransportConfigBuilder.java
```diff
@@ -11,7 +11,7 @@
  * This builder provides a fluent API for configuring the JSON-RPC transport protocol.
  * All configuration options are optional - if not specified, sensible defaults are used:
  * <ul>
- *   <li><b>HTTP client:</b> {@link JdkA2AHttpClient} (JDK's built-in HTTP client)</li>
+ *   <li><b>HTTP client:</b> Auto-selected via {@link A2AHttpClientFactory} (prefers Vert.x, falls back to JDK)</li>
```
kabir, line 14: Is 'prefers Vert.x' really needed? AIUI this depends on if it is on the classpath or not?

reference/rest/src/test/java/io/a2a/server/rest/quarkus/QuarkusA2ARestTest.java
```diff
@@ -33,7 +39,7 @@ protected String getTransportUrl() {
 
     @Override
     protected void configureTransport(ClientBuilder builder) {
-        builder.withTransport(RestTransport.class, new RestTransportConfigBuilder());
+        builder.withTransport(RestTransport.class, new RestTransportConfigBuilder().httpClient(new VertxA2AHttpClient(vertx)));
```
kabir: @ehsavoie Is the thinking here that JSONRPC will use the JDK one, and the REST ones the Vert.x ones? 

If so that needs documenting.

kabir (changes requested):
In addition to my comment about REST using vert.x client, and JSON-RPC using JDK client, (assuming it is correct)I think it would be good to have at least some sanity tests for the inverse case somewhere.

i.e. some streaming and non-streaming calls with REST with JDK client, and the same for JSON-RPC and vert.x?

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransportConfigBuilder.java
ehsavoie: if it is on the classpath and vertx lib also it will use vertx

reference/rest/src/test/java/io/a2a/server/rest/quarkus/QuarkusA2ARestTest.java
ehsavoie: I think it is best ot make rest  and jsonrpc tests parametrizable to test both clients

PR #589, opened 2026-01-15
fix: add missing META-INF/beans.xml to microprofile-config integration

### LINKED ISSUE #587: [Bug]: Microprofile config integration is missing the beans.xml file
### What happened?

Looks like the microprofile-config integration is missing the `META-INF/beans.xml` marker file, which means that the configuration is not getting loaded as an alternative bean in Quarkus applications.

Is this intentional? or was this a miss?

Noticed this issue with `0.3.3` ... and looks like the `main` branch doesn't have it either.

For now, I am pushing my own bean that simply extends from `MicroProfileConfigProvider`

### Relevant log output

```shell

```

### Code of Conduct

roguexz:
The microprofile-config integration was missing the beans.xml marker file, which prevented the configuration from being loaded as an alternative bean in Quarkus applications. This file is required for CDI bean discovery.

Fixes #587

### DIFF
--- a/integrations/microprofile-config/src/main/resources/META-INF/beans.xml
+++ b/integrations/microprofile-config/src/main/resources/META-INF/beans.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee 
+                           https://jakarta.ee/xml/ns/jakartaee/beans_4_0.xsd">
+</beans>


kabir:
Thanks @roguexz - while an empty beans.xml is fine, to stay consistent with the other ones would you be able to add content similar to the other ones in the project? e.g:
````
<?xml version="1.0" encoding="UTF-8"?>
<beans xmlns="https://jakarta.ee/xml/ns/jakartaee"
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xsi:schemaLocation="https://jakarta.ee/xml/ns/jakartaee 
                           https://jakarta.ee/xml/ns/jakartaee/beans_4_0.xsd">
</beans>
````

As this is an SDK that can be run anywhere, we don't know how lenient the target environment will be :-)

roguexz:
Sure thing. Will do in a few hours when I get back home.

roguexz:
Done. Updated and rebased the branch.

PR #596, opened 2026-01-16
docs: add AGENTS.md coding guide

### LINKED ISSUE #595: docs: add AGENTS.md coding guide
Add an AGENTS.md to help with vibe coding.

ehsavoie:
Provides examples and guidelines for AI-assisted coding.

Fixes #595 🦕

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,592 @@
+# AGENTS.md - AI Agent Guide for A2A Java SDK
+
+This document provides essential context for AI agents working with the A2A Java SDK codebase.
+
+## Project Overview
+
+**A2A Java SDK** is a comprehensive Java implementation of the [Agent2Agent (A2A) Protocol](https://a2a-protocol.org/), enabling Java applications to run as A2A servers and communicate with other A2A agents.
+
+- **Language**: Java 17+
+- **Build Tool**: Maven 3.x
+- **Primary Framework**: Quarkus (for reference implementations)
+- **License**: Apache 2.0
+- **Repository**: https://github.com/a2aproject/a2a-java
+
+## Quick Start Guide
+
+This 5-minute guide shows you how to create your first A2A server and client using the helloworld example.
+
+### Prerequisites
+
+- Java 17 or higher
+- Maven 3.6+
+- An IDE or text editor
+
+### Build the SDK
+
+First, clone and build the A2A Java SDK:
+
+```bash
+git clone https://github.com/a2aproject/a2a-java.git
+cd a2a-java
+mvn clean install -DskipTests=true
+```
+
+### Create a Simple A2A Server
+
+The minimal A2A server requires two components:
+
+**1. AgentCard Producer** (`AgentCardProducer.java`):
+
+```java
+@ApplicationScoped
+public class AgentCardProducer {
+    @Produces
+    @PublicAgentCard
+    public AgentCard agentCard() {
+        return AgentCard.builder()
+            .name("Hello World Agent")
+            .description("Just a hello world agent")
+            .supportedInterfaces(Collections.singletonList(
+                new AgentInterface("jsonrpc", "http://localhost:9999")))
+            .version("1.0.0")
+            .capabilities(AgentCapabilities.builder()
+                .streaming(true)
+                .build())
+            .protocolVersion(AgentCard.CURRENT_PROTOCOL_VERSION)
+            .build();
+    }
+}
+```
+
+**2. AgentExecutor Implementation** (`AgentExecutorProducer.java`):
+
+```java
+@ApplicationScoped
+public class AgentExecutorProducer {
+    @Produces
+    public AgentExecutor agentExecutor() {
+        return new AgentExecutor() {
+            @Override
+            public void execute(RequestContext context, EventQueue eventQueue)
+                    throws A2AError {
+                eventQueue.enqueueEvent(A2A.toAgentMessage("Hello World"));
+            }
+
+            @Override
+            public void cancel(RequestContext context, EventQueue eventQueue)
+                    throws A2AError {
+                throw new UnsupportedOperationError();
+            }
+        };
+    }
+}
+```
+
+**Run the Server**:
+
+```bash
+cd examples/helloworld/server
+mvn quarkus:dev
+```
+
+The server will start on `http://localhost:9999`.
+
+### Create a Simple A2A Client
+
+**Minimal Client Example** (`HelloWorldClient.java`):
+
+```java
+public class HelloWorldClient {
+    public static void main(String[] args) throws Exception {
+        // 1. Fetch the agent card from the server
+        AgentCard agentCard = new A2ACardResolver("http://localhost:9999")
+            .getAgentCard();
+
+        // 2. Create event consumer to handle responses
+        CompletableFuture<String> response = new CompletableFuture<>();
+        List<BiConsumer<ClientEvent, AgentCard>> consumers = new ArrayList<>();
+        consumers.add((event, card) -> {
+            if (event instanceof MessageEvent messageEvent) {
+                Message msg = messageEvent.getMessage();
+                response.complete(extractText(msg));
+            }
+        });
+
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,479 @@
+# AGENTS.md - AI Agent Guide for A2A Java SDK
+
+This document provides essential context for AI agents working with the A2A Java SDK codebase.
+
+## Project Overview
+
+**A2A Java SDK** is a comprehensive Java implementation of the [Agent2Agent (A2A) Protocol](https://a2a-protocol.org/), enabling Java applications to run as A2A servers and communicate with other A2A agents.
+
+- **Language**: Java 17+
+- **Build Tool**: Maven
```
jmesnil: Maven 3.x maybe?

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+- **Primary Framework**: Quarkus (for reference implementations)
```
jmesnil, line 11: We should instead state that the SDK is framework agnostic and provides a reference server based on Quarkus.
We don't want AI to bring Quarkus code in the core of the SDK

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+- **License**: Apache 2.0
+- **Repository**: https://github.com/a2aproject/a2a-java
+
+## Quick Start Guide
```
jmesnil, line 15: is that required in the AGENTS.md? It seems to be user doc (that should be in the `README`)

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+- Try different transports (gRPC, REST) instead of JSON-RPC
+- Add task management with `TaskUpdater`
+- Implement streaming responses
+- Add skills and capabilities to your AgentCard
+
+See the "Common Development Tasks" section below for more advanced patterns.
+
+## Architecture
+
+### Multi-Module Maven Project Structure
```
jmesnil, line 179: I'm not sure this file should contain a snapshot of the layout. The AI can figure it out.
It'd be better to state that the project uses a multi-module mven project structure and separates its different components based on their targets (server-side, client-side), transport and goals (eg reference server for tck)

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+│   ├── task-store-database-jpa/
+│   ├── push-notification-config-store-database-jpa/
+│   └── queue-manager-replicated/
+├── boms/                   # Bill of Materials for dependency management
+├── examples/               # Example applications
+├── tck/                    # Technology Compatibility Kit
+└── tests/                  # Test utilities
+```
+
+### Key Concepts
```
jmesnil, line 234: let's point this file to https://a2a-protocol.org/latest/specification/ instead?

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+
+# Skip tests
+mvn clean install -DskipTests=true
+
+# Build specific module
+cd <module-directory>
+mvn clean install
+```
+
+### Code Quality Standards
```
jmesnil, line 267: In terms of quality standards, we should let the agents know how the code should be written.

Emphasise code reuse and consistency. Be succinct in comments (they must add something to the code), etc.

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+- `SECURITY.md`: Security policy
+- `examples/*/README.md`: Example-specific documentation
+
+### Configuration
+- `pom.xml`: Root Maven configuration
+- `*/pom.xml`: Module-specific configurations
+- `META-INF/a2a-defaults.properties`: Default configuration values
+- `.github/workflows/`: CI/CD pipelines
+
+### Code Generation
```
jmesnil, line 385: that one should be further expanded to let the Agents know that a2a.proto is the source of truth and must never be updated.

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+- `spec-grpc/src/main/proto/a2a.proto`: gRPC protocol definition
+- Regenerate with: `mvn clean install -Dskip.protobuf.generate=false`
+
+## Module Dependencies
```
jmesnil, line 390: the Agent can understand the module dependencies. 
It would be more useful to speficy the intent of the dependencies.
transport modules depends on spec , etc.

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+- `boms/sdk`: Core SDK dependencies
+- `boms/reference`: Reference implementation dependencies
+- `boms/extras`: Optional extras dependencies
+- `boms/test-utils`: Test utilities
+
+## Common Patterns in Codebase
+
+### Error Handling
+```java
+// Custom errors extend JSONRPCError
```
jmesnil: that's way to specific. Depending on the modules, context, etc. the exception can extend other errors

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+throw new TaskNotFoundError("Task not found: " + taskId);
+throw new TaskNotCancelableError();
+```
+
+### Async Operations
```
jmesnil, line 423: this is too specific and would give false information.
Using Mutiny Zero for reactive streams makes sense for modules that depends on Quarkus stack only (afaict)

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+
+### Debug Tips
+
+- Enable debug logging: Set `quarkus.log.level=DEBUG` in `application.properties`
+- Use `maven.test.redirectTestOutputToFile=false` to see test output in console
+- Check `target/surefire-reports/` for test failure details
+
+## Version Information
+
+- **Current Version**: 1.0.0.Alpha1-SNAPSHOT
```
jmesnil: this will be out of date, instead we should tell the agent that the current version can be found in the root pom.xml
the Quarkus version is specified by the `quarkus.platform.version` in the root pom, etc.

AGENTS.md
```diff
@@ -0,0 +1,479 @@
...
+
+## Contributing
+
+See `CONTRIBUTING.md` for:
+- Fork and clone process
+- Branch naming conventions
+- PR requirements (tests, documentation, commit messages)
+- Code review process
+
+## Notes for AI Agents
```
jmesnil, line 575: Isn't the whole content of this file meant for agents?
This should complement the `README` not duplicate or replace its content

jmesnil (commented):
I'm not a big agent user so take everything I commented with a grain of salt.

But this file looks like a SNAPSHOT of the content of this repo at this current time instead of explanation of what is the intent of that repo

AGENTS.md
ehsavoie: The goal is for the coding agent to get those elements, having a url would mean it needs to fetch it which requires additional permissions

AGENTS.md
ehsavoie: Claude seemed to think so

AGENTS.md
ehsavoie: I've asked the agent to follow KISS and DRY principles

AGENTS.md
ehsavoie: Added

AGENTS.md
ehsavoie: I'm not sure there

kabir:
@jmesnil @ehsavoie Should we merge this? We can always regenerate it from the most recent changes

AGENTS.md
```diff
@@ -0,0 +1,592 @@
...
+        return AgentCard.builder()
+            .name("Hello World Agent")
+            .description("Just a hello world agent")
+            .supportedInterfaces(Collections.singletonList(
+                new AgentInterface("jsonrpc", "http://localhost:9999")))
+            .version("1.0.0")
+            .capabilities(AgentCapabilities.builder()
+                .streaming(true)
+                .build())
+            .protocolVersion(AgentCard.CURRENT_PROTOCOL_VERSION)
```
CokeLee777: I think this sample code should be changed.
ref: #691

jmesnil:
superseded by #705

Issue #597, opened 2026-01-16
[Chore]:  Centralise page token parsing


kabir:
See https://github.com/a2aproject/a2a-java/pull/460#discussion_r2699089035.

As we also have PageTokens for the PushNotificationConfigs, we should do it all at once (and other places if there are any)

PR #601, opened 2026-01-19
chore: Centralise PageToken parsing

kabir:
fixes: #597


ehsavoie (approved):
A unit test for the parsing would have been great.

PR #606, opened 2026-01-22
feat!: update a2a.proto 0572953

jmesnil:
based on commit https://github.com/a2aproject/A2A/commit/057295311b8ddda63bdda56c82a694c76d307e37

spec-grpc module has been regenerated from the updated a2a.proto

Key changes:

* Agent card
  * extendedAgentCard is not specified by capabilities (instead of `supportsExtendedAgentCard` field)
  * protocolVersions is a list (instead of a single `protocolVersion`)
* SendMessage
  * `msg` field is now named `message`
* ListTask
  * timestamp is now named statusTimestampAfter instead of `lastUpdatedAfter`
* OAuthFlows
  * `deviceCode` has been added, `implicit` & `password` have been removed
* AuthorizationCodeOAuthFlow
  * new `pkce_required` boolean field

### DIFF
--- a/client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcErrorMapper.java
+++ b/client/transport/grpc/src/main/java/io/a2a/client/transport/grpc/GrpcErrorMapper.java
@@ -3,7 +3,7 @@ package io.a2a.client.transport.grpc;
 import io.a2a.common.A2AErrorMessages;
 import io.a2a.spec.A2AClientException;
 import io.a2a.spec.ContentTypeNotSupportedError;
-import io.a2a.spec.ExtendedCardNotConfiguredError;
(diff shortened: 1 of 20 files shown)


jmesnil:
> https://github.com/jmesnil/a2a-java/blob/48d574464052bd990650d4e445751e695203a16d/spec/src/main/java/io/a2a/spec/ListTaskPushNotificationConfigParams.java hasn't been updated to have the pageSize and token

There is no change regarding them in the updated a2a.proto.
The `ListTaskPushNotificationConfigParams` record has them: 

https://github.com/jmesnil/a2a-java/blob/48d574464052bd990650d4e445751e695203a16d/spec/src/main/java/io/a2a/spec/ListTaskPushNotificationConfigParams.java#L16

What would be missing?

ehsavoie:
@jmesnil sorry, I read the code too quickly.

PR #610, opened 2026-01-23
feat: Add an optional tenant parameter to the getExtendedAgentCard operation

### LINKED ISSUE #608: [Bug]: GetExtendedAgentCard request do not support the optional tenant parameter
### What happened?

The client must provide an optional `tenant` field when calling the method to get the extended agent card:

https://github.com/a2aproject/A2A/blob/92e6fb712ebeddd732a0a124b95d5ff8db4e0d41/specification/grpc/a2a.proto#L849

### Code of Conduct

jmesnil:
Fixes #608 🦕

### DIFF
--- a/client/base/src/main/java/io/a2a/client/AbstractClient.java
+++ b/client/base/src/main/java/io/a2a/client/AbstractClient.java
@@ -390,7 +390,18 @@ public abstract class AbstractClient implements AutoCloseable {
      * @throws A2AClientException if retrieving the extended agent card fails for any reason
      */
     public AgentCard getExtendedAgentCard() throws A2AClientException {
-        return getExtendedAgentCard(null);
+        return getExtendedAgentCard(null, null);
+    }
+
+    /**
+     * Retrieve the extended AgentCard.
+     *
+     * @param tenant Optional tenant
+     * @return the extended AgentCard
+     * @throws A2AClientException if retrieving the extended agent card fails for any reason
+     */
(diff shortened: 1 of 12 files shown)


jmesnil:
@kabir @ehsavoie I've followed the same patterns than the other operations to take into account the optional `tenant` parameter.
However, I'm not sure if what we have for gRPC is correct.
For JsonRPC & REST, we append the tenant to the URL before sending the HTTP request.
For gRPC, we do nothing? From the protobuf, it seems that we should change the name of the operation (eg https://github.com/a2aproject/A2A/blob/92e6fb712ebeddd732a0a124b95d5ff8db4e0d41/specification/grpc/a2a.proto#L121) but I'm not sure were to do that...

ehsavoie:
@jmesnil afaik for grpc the tenant is a parameter https://github.com/a2aproject/A2A/blob/92e6fb712ebeddd732a0a124b95d5ff8db4e0d41/specification/grpc/a2a.proto#L849

jmesnil:
ok I see. For gRPC, tenant is just a parameter (as opposed to JSON-RPC & REST where the tenant is put into the endpoint URL).
Then it's up to the server implementation to support it.... Which is maybe something to bring to the spec? Can a server ignore the "tenant" parameter? Should the capabilities of the agent card specify that it supports multitenancy?

kabir (approved):
It looks fine to me but @ehsavoie is more familiar with this tenant stuff :-)

PR #611, opened 2026-01-26
feat: Implement MainEventBus architecture for event queue processing

kabir:
Introduces centralized event processing with single background thread to
guarantee event persistence before client visibility and eliminate race
conditions in concurrent task updates.

Key Changes:
- MainEventBus: Central LinkedBlockingDeque for all events
- MainEventBusProcessor: Single background thread ensuring serial processing
  (TaskStore.save() -> PushNotificationSender.send() -> distributeToChildren())
- Two-level queue cleanup protection via TaskStateProvider.isTaskFinalized()
  to prevent premature cleanup for fire-and-forget tasks
- Deterministic blocking calls: waitForTaskFinalization() ensures TaskStore
  persistence completes before returning to client
- Streaming closure: agentCompleted flag via EnhancedRunnable.DoneCallback
  for graceful drain when agent completes
- SseFormatter utility: Framework-agnostic SSE formatting in server-common
- Executor pool improvements: Bounded EventConsumerExecutor pool (size 15)
  prevents exhaustion during high concurrency

Null TaskId Support:
- QueueManager.switchKey(): Atomic key switching from temp to real task ID
- Non-streaming path now handles null taskId like streaming (generates temp UUID)
- Cloud deployment test reverted to use A2A.toUserMessage() (valid null taskId case)
- No duplicate queue map entries, clean queue lifecycle management

Additional Fixes:
- Blocking calls always wait for task finalization (not just final events)
- RequestContext uses message.taskId() for new tasks when task is null
- PostgreSQL pod deployment: Wait loop for pod creation before readiness check
- Javadoc: Remove invalid <p> tags around <pre> blocks
- ServerCallContext: EventConsumer cancellation via closeHandler for graceful disconnect
- Gemini feedback improvements

Architecture Impact:
- All events flow through MainEventBus before distribution to ChildQueues
- Clients never see unpersisted events (persistence-before-distribution)
- Fire-and-forget tasks supported: queues stay open for non-final states
- Late resubscription enabled: queues persist until task finalization
- Null taskId messages supported: temp IDs transition to real IDs via switchKey
- Test synchronization: MainEventBusProcessorCallback for deterministic testing

### DIFF
--- a/extras/queue-manager-replicated/core/src/test/java/io/a2a/extras/queuemanager/replicated/core/ReplicatedQueueManagerTest.java
+++ b/extras/queue-manager-replicated/core/src/test/java/io/a2a/extras/queuemanager/replicated/core/ReplicatedQueueManagerTest.java
@@ -312,6 +312,14 @@ class ReplicatedQueueManagerTest {
         ExecutorService executor = Executors.newFixedThreadPool(numThreads);
         CountDownLatch startLatch = new CountDownLatch(1);
         CountDownLatch doneLatch = new CountDownLatch(numThreads);
+        
+        // Use CyclicBarrier for better thread synchronization
+        // This ensures all threads start their work at approximately the same time
+        java.util.concurrent.CyclicBarrier barrier = new java.util.concurrent.CyclicBarrier(numThreads);
+
+        // Track processed events for better diagnostics on failure
+        java.util.concurrent.CopyOnWriteArrayList<io.a2a.spec.Event> processedEvents = 
+            new java.util.concurrent.CopyOnWriteArrayList<>();
 
         // Set up callback to wait for ALL events to be processed by MainEventBusProcessor
         // Must wait for all 50 events (25 normal + 25 replicated) to ensure all normal events
@@ -320,6 +328,7 @@ class ReplicatedQueueManagerTest {
         mainEventBusProcessor.setCallback(new io.a2a.server.events.MainEventBusProcessorCallback() {
             @Override
             public void onEventProcessed(String tid, io.a2a.spec.Event event) {
+                processedEvents.add(event);
                 processingLatch.countDown();
             }
 
@@ -335,6 +344,7 @@ class ReplicatedQueueManagerTest {
             executor.submit(() -> {
                 try {
                     startLatch.await();
+                    barrier.await(); // Synchronize thread starts for better interleaving
                     for (int j = 0; j < eventsPerThread; j++) {
                         TaskStatusUpdateEvent event = TaskStatusUpdateEvent.builder()
                                 .taskId(taskId)  // Use same taskId as queue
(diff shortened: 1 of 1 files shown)


extras/queue-manager-replicated/core/src/main/java/io/a2a/extras/queuemanager/replicated/core/ReplicatedQueueManager.java
```diff
@@ -87,48 +90,86 @@ public void awaitQueuePollerStart(EventQueue eventQueue) throws InterruptedExcep
...
+                LOGGER.warn(
+                        "MainQueue not found for task {}, cannot enqueue replicated event. This may happen if the queue was already cleaned up.",
+                        replicatedEvent.getTaskId());
             }
         } finally {
-            // Close the temporary ChildQueue to prevent leaks
-            // The MainQueue remains open for other consumers
-            childQueue.close();
+            if (childQueue != null) {
+                childQueue.close();  // Close the ChildQueue we created (not MainQueue!)
```
ehsavoie: if close() throws an exception it will 'hide' previous exception

client/transport/rest/src/main/java/io/a2a/client/transport/rest/sse/RestSSEEventListener.java
```diff
@@ -62,6 +65,23 @@ private void handleMessage(StreamResponse response) {
             }
         }
         eventHandler.accept(event);
+
+        // Client-side auto-close on final events to prevent connection leaks
+        // Handles both TaskStatusUpdateEvent and Task objects with final states
+        // This covers late subscriptions to completed tasks and ensures no connection leaks
+        boolean shouldClose = false;
```
ehsavoie, line 72: This seems to be duplicate of JSONRPC own update

kabir: @ehsavoie Not sure what you mean here? Won't the client be using either this or JSONRPC (or gRPC)?

extras/queue-manager-replicated/core/src/main/java/io/a2a/extras/queuemanager/replicated/core/ReplicatedQueueManager.java
kabir: @ehsavoie the closing is safe AFAICT, and this is internal code used when receiving replication events. But I've added a catch and debug logging just in case

client/transport/rest/src/main/java/io/a2a/client/transport/rest/sse/RestSSEEventListener.java
kabir: Tracked in https://github.com/a2aproject/a2a-java/issues/623

PR #613, opened 2026-01-26
chore: Managing tenant in JSONRPC.

### LINKED ISSUE #612: [Bug]: JSONRPC doesn't get the tenant from the request path
### What happened?

The tenant may come from the request path of the JSONRPC request, we need to handle that case.

### Relevant log output

```shell

```

### Code of Conduct

ehsavoie:
If the tenant is not defined in the JSONRPC payload, we use the one from the request path.

Fixes #612 🦕

### DIFF
--- a/reference/jsonrpc/src/main/java/io/a2a/server/apps/quarkus/A2AServerRoutes.java
+++ b/reference/jsonrpc/src/main/java/io/a2a/server/apps/quarkus/A2AServerRoutes.java
@@ -2,6 +2,7 @@ package io.a2a.server.apps.quarkus;
 
 import static io.a2a.transport.jsonrpc.context.JSONRPCContextKeys.HEADERS_KEY;
 import static io.a2a.transport.jsonrpc.context.JSONRPCContextKeys.METHOD_NAME_KEY;
+import static io.a2a.transport.jsonrpc.context.JSONRPCContextKeys.TENANT_KEY;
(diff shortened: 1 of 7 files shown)


jmesnil:
I don't really get how multinenancy is supposed to work in A2A (even after reading https://github.com/a2aproject/A2A/issues/1273) but the resolution of the tenant from the url path looks very suspicious. 
What about having a supported interface like this:

```json
{
  "url": "https://my.example.com/a2a",
  "protocolBinding": "REST",
  "tenant": "/project/123/agent/345"
}
```

ehsavoie:
> I don't really get how multinenancy is supposed to work in A2A (even after reading [a2aproject/A2A#1273](https://github.com/a2aproject/A2A/issues/1273)) but the resolution of the tenant from the url path looks very suspicious. What about having a supported interface like this:
> 
> ```json
> {
>   "url": "https://my.example.com/a2a",
>   "protocolBinding": "REST",
>   "tenant": "/project/123/agent/345"
> }
> ```

My understanding is that the agent should know how to remove the a2a part if the tenant is something it needs to handle

jmesnil:
> My understanding is that the agent should know how to remove the a2a part if the tenant is something it needs to handle

Ok, the code I pointed out is in the reference bits which are a simple implementation of an agent. If another agent wants to support multitenancy with different approaches, it's up to them...

kabir (approved):

PR #638, opened 2026-02-10
chore: Migrate spec module to JSpecify annotations for null-safety

ehsavoie:
- Add @NullMarked to io.a2a.spec package
- Apply @Nullable annotations to optional fields and parameters
- Move builder validation to build() methods using Assert.checkNotNullParam()
- Fix null-safety issues in RequestContext ID generation
- Add null checks for push notification configuration
- Simplify REST route handling for push notification configs
- Update tests to provide required MessageSendConfiguration


spec/src/main/java/io/a2a/spec/AgentCard.java
```diff
@@ -177,9 +178,9 @@ private Builder(AgentCard card) {
             this.version = card.version;
             this.documentationUrl = card.documentationUrl;
             this.capabilities = card.capabilities;
-            this.defaultInputModes = card.defaultInputModes != null ? new ArrayList<>(card.defaultInputModes) : null;
-            this.defaultOutputModes = card.defaultOutputModes != null ? new ArrayList<>(card.defaultOutputModes) : null;
-            this.skills = card.skills != null ? new ArrayList<>(card.skills) : null;
+            this.defaultInputModes = card.defaultInputModes != null ? new ArrayList<>(card.defaultInputModes) : Collections.emptyList();
```
kabir, line 181: Should we use null still if the original was null? 
Does emptyList() instead of null change the behaviour?

spec/src/main/java/io/a2a/spec/AgentCard.java
```diff
@@ -177,9 +178,9 @@ private Builder(AgentCard card) {
...
+            this.defaultOutputModes = card.defaultOutputModes != null ? new ArrayList<>(card.defaultOutputModes) : Collections.emptyList();
+            this.skills = card.skills != null ? new ArrayList<>(card.skills) : Collections.emptyList();
```
kabir, line 183: Same null vs emptyList() questions

kabir (commented):
One small question

spec/src/main/java/io/a2a/spec/AgentCard.java
ehsavoie: I don't think it would have a big impact

server-common/src/main/java/io/a2a/server/tasks/TaskUpdater.java
```diff
@@ -0,0 +1,191 @@
...
+import io.a2a.spec.Message;
+import io.a2a.spec.Part;
+import io.a2a.spec.TaskArtifactUpdateEvent;
+import io.a2a.spec.TaskState;
+import io.a2a.spec.TaskStatus;
+import io.a2a.spec.TaskStatusUpdateEvent;
+import io.a2a.util.Assert;
+import org.jspecify.annotations.Nullable;
+
+public class TaskUpdater {
```
kabir: This class should no longer be used

server-common/src/main/java/io/a2a/server/agentexecution/RequestContext.java
```diff
@@ -246,9 +244,9 @@ public void attachRelatedTask(Task task) {
         relatedTasks.add(task);
     }
 
-    private void checkOrGenerateTaskId() {
+    private @Nullable String checkOrGenerateTaskId() {
```
kabir: It looks like this will always be set?

server-common/src/main/java/io/a2a/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -626,7 +627,7 @@ public Flow.Publisher<StreamingEventKind> onMessageSendStream(
         // Store push notification config SYNCHRONOUSLY for new tasks before agent starts
         // This ensures config is available when MainEventBusProcessor sends push notifications
         // For existing tasks, config is stored in initMessageSend()
-        if (mss.task() == null && shouldAddPushInfo(params)) {
+        if (mss.task() == null && pushConfigStore != null && params.configuration() != null && params.configuration().pushNotificationConfig() != null) {
```
kabir: Why are you removing the method and duplicating the check? :-)

PR #642, opened 2026-02-11
feat: Make RequestContext taskId and contextId nullable.

kabir:
This is possible by moving the generation of IDs when not set into the builder, and making sure the builder is used.

### DIFF
--- a/server-common/src/main/java/io/a2a/server/agentexecution/RequestContext.java
+++ b/server-common/src/main/java/io/a2a/server/agentexecution/RequestContext.java
@@ -395,8 +395,12 @@ public class RequestContext {
                         .contextId(finalContextId)
                         .build();
                 // Preserve all original fields including tenant
-                finalParams = new MessageSendParams(updatedMessage,
-                        params.configuration(), params.metadata(), params.tenant());
+                finalParams = MessageSendParams.builder()
+                        .message(updatedMessage)
+                        .configuration(params.configuration())
+                        .metadata(params.metadata())
+                        .tenant(params.tenant())
+                        .build();
             }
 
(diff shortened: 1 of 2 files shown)


server-common/src/main/java/io/a2a/server/agentexecution/RequestContext.java
```diff
@@ -357,8 +356,52 @@ public Builder setServerCallContext(@Nullable ServerCallContext serverCallContex
...
+
+            // 5. Update params if message needs to be updated with final IDs
+            MessageSendParams finalParams = params;
+            if (params != null && (!finalTaskId.equals(messageTaskId) || !finalContextId.equals(messageContextId))) {
+                Message updatedMessage = Message.builder(params.message())
+                        .taskId(finalTaskId)
+                        .contextId(finalContextId)
+                        .build();
+                // Preserve all original fields including tenant
+                finalParams = new MessageSendParams(updatedMessage,
```
ehsavoie: Shouldn't you use the builder there ?

server-common/src/main/java/io/a2a/server/tasks/AgentEmitter.java
```diff
@@ -108,7 +108,7 @@ public class AgentEmitter {
      */
     public AgentEmitter(RequestContext context, EventQueue eventQueue) {
         this.eventQueue = eventQueue;
-         this.taskId = Assert.checkNotNullParam("taskId",context.getTaskId());
+        this.taskId = Assert.checkNotNullParam("taskId",context.getTaskId());
```
ehsavoie: I don't think those Assert are required anymore

PR #657, opened 2026-02-16
fix: Why not set tenant to "" in ListTasksParams$Builder?

### LINKED ISSUE #653: Why not set tenant to "" in ListTasksParams$Builder?
Hi, I found that in ListTasksParams$Builder, the tenant field is required to be nonNull.

https://github.com/a2aproject/a2a-java/blob/7bf82c694a33ef8538c73db23d7aa42e84da3abb/spec/src/main/java/io/a2a/spec/ListTasksParams.java#L230

Whereas in MessageSendParams, the builder sets the tenant field to "" if it is null.

https://github.com/a2aproject/a2a-java/blob/7bf82c694a33ef8538c73db23d7aa42e84da3abb/spec/src/main/java/io/a2a/spec/MessageSendParams.java#L134

Why the difference? 

Considering that the constructors of ListTasksParams also set the tenant field to "" when it is not specified, would it be possible to do the same for ListTasksParams$Builder as well?

Thanks!

ehsavoie:
ListTasksParams.Builder should set the tenant to "" when it is not defined.

Fixes #653 🦕

### DIFF
--- a/spec/src/main/java/io/a2a/spec/DeleteTaskPushNotificationConfigParams.java
+++ b/spec/src/main/java/io/a2a/spec/DeleteTaskPushNotificationConfigParams.java
@@ -3,6 +3,7 @@ package io.a2a.spec;
 
 
 import io.a2a.util.Assert;
+import io.a2a.util.Utils;
 import org.jspecify.annotations.Nullable;
 
 /**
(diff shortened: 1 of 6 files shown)


jmesnil (approved):

spec/src/main/java/io/a2a/spec/DeleteTaskPushNotificationConfigParams.java
```diff
@@ -111,7 +111,7 @@ public DeleteTaskPushNotificationConfigParams build() {
             return new DeleteTaskPushNotificationConfigParams(
                     Assert.checkNotNullParam("id", id),
                     Assert.checkNotNullParam("pushNotificationConfigId", pushNotificationConfigId),
-                    Assert.checkNotNullParam("tenant", tenant));
+                    tenant != null ? tenant : "");
```
kabir: Not sure if it really is needed, but we have the `io.a2a.util.Utils.defaultIfNull()` util to help with this here and other places

kabir (commented):
Looks good. Just wondering if we should use the util

kabir (approved):

PR #658, opened 2026-02-16
feat: Add metadata support to Part types

### LINKED ISSUE #649: [Feat]: Support adding metadata to Part in 1.0.0.Alpha (regression from 0.3.x)
## Background

In version 0.3.x of the A2A Java SDK, Part included a metadata field that allowed clients to attach additional information to each part.

However, after the redesign in 1.0.0.Alpha1, the metadata field was removed from Part.

## Problem

Previously, metadata in Part (such as TextPart, DataPart, and FilePart) was useful for passing additional contextual or domain-specific information to clients.

For example:

- Custom rendering hints
- Classification tags
- Correlation identifiers
- Client-specific processing flags

### Proposal

Could you please consider reintroducing support for metadata in Part (or an equivalent extensibility mechanism)?

_No response_

### Additional context

_No response_

### Code of Conduct

ehsavoie:
Add optional metadata field to TextPart, FilePart, and DataPart records to support additional contextual information for message and artifact content parts as specified in the A2A protocol v1.0.

Changes:
  - Add metadata parameter to Part constructors (TextPart, FilePart, DataPart)
  - Update PartMapper to convert metadata between domain and gRPC proto
  - Update JsonUtil PartTypeAdapter to serialize/deserialize metadata

Fixes #649 🦕

### DIFF
--- a/spec/src/main/java/io/a2a/spec/DataPart.java
+++ b/spec/src/main/java/io/a2a/spec/DataPart.java
@@ -2,6 +2,7 @@ package io.a2a.spec;
 
 
 import io.a2a.util.Assert;
+import java.util.Map;
 import org.jspecify.annotations.Nullable;
 
 
@@ -37,11 +38,12 @@ import org.jspecify.annotations.Nullable;
  * }</pre>
  *
  * @param data the structured data (required, supports JSON objects, arrays, primitives, and null)
+ * @param metadata additional metadata for the part
  * @see Part
  * @see Message
  * @see Artifact
  */
-public record DataPart(Object data) implements Part<Object> {
+public record DataPart(Object data, @Nullable Map<String, Object> metadata) implements Part<Object> {
 
     /**
      * The JSON member name discriminator for data parts: "data".
@@ -60,50 +62,19 @@ public record DataPart(Object data) implements Part<Object> {
      * @param data the structured data (supports JSON objects, arrays, primitives, and null)
      * @throws IllegalArgumentException if data is null
      */
-    public DataPart {
+    public DataPart (Object data, @Nullable Map<String, Object> metadata) {
         Assert.checkNotNullParam("data", data);
+        this.metadata = metadata == null ? null : Map.copyOf(metadata);
+        this.data = data;
     }
 
     /**
-     * Create a new Builder
+     * Constructor.
      *
-     * @return the builder
+     * @param data the structured data (supports JSON objects, arrays, primitives, and not null)
+     * @throws IllegalArgumentException if data is null
      */
-    public static Builder builder() {
-        return new Builder();
-    }
-
-    /**
-     * Builder for constructing {@link DataPart} instances.
-     */
-    public static class Builder {
-        private @Nullable Object data;
-
-        /**
-         * Creates a new Builder with all fields unset.
-         */
-        private Builder() {
-        }
-
-        /**
-         * Sets the structured data.
-         *
-         * @param data the structured data (required, supports JSON objects, arrays, primitives, and null)
-         * @return this builder for method chaining
-         */
-        public Builder data(Object data) {
-            this.data = data;
-            return this;
-        }
-
-        /**
-         * Builds a new {@link DataPart} from the current builder state.
-         *
-         * @return a new DataPart instance
-         * @throws IllegalArgumentException if data is null
-         */
-        public DataPart build() {
-            return new DataPart(Assert.checkNotNullParam("data", data));
-        }
+    public DataPart(Object data) {
+        this(data, null);
     }
 }
--- a/spec/src/main/java/io/a2a/spec/FilePart.java
+++ b/spec/src/main/java/io/a2a/spec/FilePart.java
@@ -2,6 +2,8 @@ package io.a2a.spec;
 
 
 import io.a2a.util.Assert;
+import java.util.Map;
+import org.jspecify.annotations.Nullable;
 
 
 /**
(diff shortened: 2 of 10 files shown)


extras/push-notification-config-store-database-jpa/src/test/java/io/a2a/extras/pushnotificationconfigstore/database/jpa/JpaDatabasePushNotificationConfigStoreIntegrationTest.java
```diff
@@ -105,7 +106,7 @@ public void testJpaDatabasePushNotificationConfigStoreIntegration() throws Excep
         // Step 1: Create the task
         Message createMessage = Message.builder()
             .role(Message.Role.USER)
-            .parts(List.of(new TextPart("create"))) // Send the "create" command
+            .parts(List.of(new TextPart("create", Collections.emptyMap()))) // Send the "create" command
```
kabir: Why is it an emptyMap() in this test (and some more of the extras ones) and null everywhere else?

spec/src/main/java/io/a2a/spec/TextPart.java
```diff
@@ -20,11 +22,12 @@
  * }</pre>
  *
  * @param text the text content (required, must not be null)
+ * @param metadata additional metadata for the part
  * @see Part
  * @see Message
  * @see Artifact
  */
-public record TextPart(String text) implements Part<String> {
+public record TextPart(String text, @Nullable Map<String, Object> metadata) implements Part<String> {
```
kabir, line 30: I don't know if it will break the mappers, but perhaps an overloaded constructor here and in the other Parts would remove the need for null/emptyMap everywhere.
Then again, we have the builder, so perhaps the pattern I have been seeing is more for convenience

kabir (commented):
Needs a rebase

extras/push-notification-config-store-database-jpa/src/test/java/io/a2a/extras/pushnotificationconfigstore/database/jpa/JpaDatabasePushNotificationConfigStoreIntegrationTest.java
ehsavoie: because initially in the builder I was setting an empty map which lead to assertion failures

spec/src/main/java/io/a2a/spec/TextPart.java
jmesnil: We don't have builder for the parts, right?

+1  to add an overloaded constructor

```
    public TextPart(String text) {
        this(text, null); // or emptyMap()?
    }
```

jmesnil (changes requested):

kabir (approved):

spec/src/main/java/io/a2a/spec/DataPart.java
```diff
@@ -78,6 +91,7 @@ public static Builder builder() {
      */
     public static class Builder {
```
jmesnil: Do we really need a builder for this?

spec/src/main/java/io/a2a/spec/DataPart.java
```diff
@@ -37,11 +39,12 @@
  * }</pre>
  *
  * @param data the structured data (required, supports JSON objects, arrays, primitives, and null)
+ * @param metadata additional metadata for the part
  * @see Part
  * @see Message
  * @see Artifact
  */
-public record DataPart(Object data) implements Part<Object> {
+public record DataPart(Object data, @Nullable Map<String, Object> metadata) implements Part<Object> {
```
jmesnil, line 46: If a DataPart is a record, I think its fields should be immutable too.
The metadata should be copied to an Immutable map.
I also think that there is no good reason to allow it to be null. Either it set by the user (and make immutable in the record) or the user did not set it and an emptyMap() can be used

spec/src/main/java/io/a2a/spec/FilePart.java
```diff
@@ -55,4 +58,14 @@ public record FilePart(FileContent file) implements Part<FileContent> {
     public FilePart {
         Assert.checkNotNullParam("file", file);
     }
+
+    /**
+     * Constructor.
+     *
+     * @param file the file content (required, either FileWithBytes or FileWithUri)
+     * @throws IllegalArgumentException if data is null
```
jmesnil: no data field

spec/src/main/java/io/a2a/spec/DataPart.java
ehsavoie: The (de)serialization differs if it is null or empty which is why i would keep null as a value

jmesnil: I'm curious. How does it differ? https://github.com/a2aproject/a2a-java/pull/658/changes#diff-d5f468124317cecd0e32379b93a34f20c5c9c44c2adb5ab3415569520e36f3b7R511 treats null & an empty map the same way.
And in https://github.com/a2aproject/a2a-java/pull/658/changes#diff-d5f468124317cecd0e32379b93a34f20c5c9c44c2adb5ab3415569520e36f3b7R566, if there is no metadata in the json, it can be instantiated as an empty map instead of null.

ehsavoie: the gson will produce an empty json map instead of no field at all

jmesnil (approved):

PR #673, opened 2026-02-18
fix: Add no-arg overload for getUserInput() in RequestContext to avoid

### LINKED ISSUE #668: [Feat]: Add no-arg overload for `getUserInput()` in `RequestContext` to avoid passing null for default delimiter

In `io.a2a.server.agentexecution.RequestContext` (`a2a-java-sdk-server-common`), the current `getUserInput(String delimiter)` method uses `null` as a signal to apply a default delimiter (`"\n"`). This is an unintuitive API contract that forces callers to pass `null` explicitly to get default behavior, which is not idiomatic Java.

```java
// Current: caller must pass null to use default delimiter
context.getUserInput(null);
```

Add a no-argument overload that delegates to the existing method with the default delimiter, without changing the existing signature:

```java
public String getUserInput() {
    return getUserInput("\n");
}
```

This allows callers to write:

```java
context.getUserInput();      // default "\n" delimiter
context.getUserInput(" ");  // custom delimiter
```

Replacing the existing method entirely with a cleaner signature that removes the `null` handling — this would be a breaking change and is not preferred.

### Additional context

The existing Javadoc already acknowledges the awkwardness with the `null` example (`context.getUserInput(null); // uses "\n"`), which suggests this was recognized as non-ideal at the time of writing. Adding the no-arg overload is fully backward-compatible and requires minimal change.

I'd be happy to submit a PR for this if the proposal is accepted.

### Code of Conduct

ehsavoie:
passing null for default delimiter.

Fixes issue #668

Fixes #668 🦕

### DIFF
--- a/server-common/src/main/java/io/a2a/server/agentexecution/RequestContext.java
+++ b/server-common/src/main/java/io/a2a/server/agentexecution/RequestContext.java
@@ -225,6 +225,24 @@ public class RequestContext {
     }
 
     /**
+     * Extracts all text content from the message and joins with the newline delimiter.
+     * <p>
+     * This is a convenience method for getting text input from messages that may contain
+     * multiple text parts. Non-text parts (images, etc.) are ignored.
+     * <p>
+     * <b>Examples:</b>
+     * <pre>{@code
+     * // Join with newlines (common for multi-paragraph input)
+     * String text = context.getUserInput();
+     * }</pre>
+     *
+     * @return all text parts joined with newline, or empty string if no message
+     */
+    public String getUserInput() {
+        return getUserInput("\n");
+    }
+
+    /**
      * Extracts all text content from the message and joins with the specified delimiter.
      * <p>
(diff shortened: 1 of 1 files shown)


server-common/src/main/java/io/a2a/server/agentexecution/RequestContext.java
```diff
@@ -245,14 +263,15 @@ public List<Task> getRelatedTasks() {
...
         if (params == null) {
             return "";
         }
+        String delim = delimiter;
         if (delimiter == null) {
-            delimiter = "\n";
+            delim = "\n";
         }
-        return getMessageText(params.message(), delimiter);
+        return getMessageText(params.message(), delim);
```
CokeLee777: Small suggestion. How about using a ternary operator here? It might make the code simpler.

`return getMessageText(params.message(), (delimiter == null) ? "\n" : delimiter);`

PR #692, opened 2026-03-01
fix!: Remove includingDefaultValueFields() from JsonFormat.printer() in JSONRPCUtils

### LINKED ISSUE #689: [Bug]: RestHandler#sendMessage returns part with both "text" and "filename" fields
### What happened?

Hello!
I am currently using Alpha3 and ran into this problem:

I called RestHandler#sendMessage and expected to create a task with one TextPart.
But we get a part object like this:

```json
{
  "tasks": [
  {
    "xxx": "xxx",
    "parts": [
      {
        "text": "abc",
        "metadata": {},
        "filename": "",
        "mediaType": ""
      }
    ]
  }
]
}
```

I believe the filename field is not supposed to be there.

I also found RestHandler use grpc models/functions internally, which might be the problem: when it performs proto deserialization, it added the filename to the part model.

### Relevant log output

```shell

```

### Code of Conduct

CokeLee777:
# Description

Remove the `includingDefaultValueFields()` call from `JsonFormat.printer()` in `toJsonRPCRequest` and `toJsonRPCResultResponse` methods to prevent unset protobuf fields (e.g. `filename`, `mediaType`, `metadata`) from being emitted as default values in the serialized JSON output.

Add tests to verify that a `TextPart` with no metadata serializes with only the `text` field present, and that optional fields such as `filename`, `mediaType`, and `metadata` are excluded when not explicitly set.

Fixes #689

### DIFF
--- a/spec-grpc/src/main/java/io/a2a/grpc/utils/JSONRPCUtils.java
+++ b/spec-grpc/src/main/java/io/a2a/grpc/utils/JSONRPCUtils.java
@@ -576,7 +576,7 @@ public class JSONRPCUtils {
                 output.name("method").value(method);
             }
             if (payload != null) {
-                String resultValue = JsonFormat.printer().includingDefaultValueFields().omittingInsignificantWhitespace().print(payload);
+                String resultValue = JsonFormat.printer().omittingInsignificantWhitespace().print(payload);
                 output.name("params").jsonValue(resultValue);
             }
             output.endObject();
@@ -599,7 +599,7 @@ public class JSONRPCUtils {
                     output.name("id").value(number.longValue());
                 }
             }
-            String resultValue = JsonFormat.printer().includingDefaultValueFields().omittingInsignificantWhitespace().print(builder);
+            String resultValue = JsonFormat.printer().omittingInsignificantWhitespace().print(builder);
             output.name("result").jsonValue(resultValue);
             output.endObject();
             return result.toString();
--- a/spec-grpc/src/test/java/io/a2a/grpc/utils/JSONRPCUtilsTest.java
+++ b/spec-grpc/src/test/java/io/a2a/grpc/utils/JSONRPCUtilsTest.java
@@ -1,32 +1,23 @@
 package io.a2a.grpc.utils;
 
-import static io.a2a.grpc.utils.JSONRPCUtils.ERROR_MESSAGE;
-import static io.a2a.spec.A2AMethods.GET_TASK_PUSH_NOTIFICATION_CONFIG_METHOD;
-import static io.a2a.spec.A2AMethods.SET_TASK_PUSH_NOTIFICATION_CONFIG_METHOD;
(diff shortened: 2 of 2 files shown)


ehsavoie:
This breaks the TCK:

__________________________________________________________________________________________________________________ TestBasicListing.test_list_tasks_empty_when_none_exist __________________________________________________________________________________________________________________
tests/mandatory/protocol/test_tasks_list_method.py:163: in test_list_tasks_empty_when_none_exist
    assert result["tasks"] == [], "Tasks array should be empty for non-existent context"
           ^^^^^^^^^^^^^^^
E   KeyError: 'tasks'
_________________________________________________________________________________________________________________________ TestPagination.test_last_page_detection __________________________________________________________________________________________________________________________
tests/mandatory/protocol/test_tasks_list_method.py:528: in test_last_page_detection
    assert result["nextPageToken"] == "", \
           ^^^^^^^^^^^^^^^^^^^^^^^
E   KeyError: 'nextPageToken'


@jmesnil should we update the TCK or refuse this global change ?

jmesnil:
According to the proto but definition:

> // Array of tasks matching the specified criteria.
>  repeated Task tasks = 1 [(google.api.field_behavior) = REQUIRED];
>  // A token to retrieve the next page of results, or empty if there are no more results in the list.
>  string next_page_token = 2 [(google.api.field_behavior) = REQUIRED];

So the `tasks` should be present (and empty) and the `next_page_token` should be present (and potentially empty).

It seems the a2a-java shouldn't "brute-force" the fields omission from the JSON payload

CokeLee777:
@jmesnil 
Thank you for the feedback, I understand that removing `includingDefaultValueFields()` globally breaks the TCK for REQUIRED fields like `tasks` and `nextPageToken`.

However, I'd like to point out a related issue that exists on the current `main` branch:

When a message is serialized via `JSONRPCUtils.toJsonRPCRequest()` with `includingDefaultValueFields()`, a `TextPart` is serialized as:

```json
{"text":"hello","filename":"","mediaType":"","metadata":{}}
```

When this JSON is then parsed by `PartTypeAdapter` in `jsonrpc-common`, it fails because the adapter enforces a strict key count check (`keys.size() > 2`), and the extra default-value fields (`filename`, `mediaType`) cause a `JsonSyntaxException`.

So with the current `main`, there is a compatibility issue between the gRPC serialization and the JSON-RPC deserialization for `Part` objects.

Would a more targeted approach work here? For example:

1. Using `includingDefaultValueFields(Set<FieldDescriptor>)` to selectively include defaults only for REQUIRED fields (like `tasks`, `nextPageToken`)
2. Or making `PartTypeAdapter` more lenient by filtering for known keys instead of enforcing strict key count

I'm happy to adjust the PR to whichever approach you prefer. What do you think?

ehsavoie:
This should be superseded by https://github.com/a2aproject/a2a-java/pull/700

PR #699, opened 2026-03-04
feat: Add in-process context propagation for OpenTelemetry agent execution

### LINKED ISSUE #698: [Feat]: Propagate open telemetry context to agent execution for consolidated tracing

Hello, 

A2A Java SDK supports server OpenTelemetry thanks to [OpenTelemetryRequestHandlerDecorator](https://github.com/a2aproject/a2a-java/blob/main/extras/opentelemetry/server/src/main/java/io/a2a/extras/opentelemetry/OpenTelemetryRequestHandlerDecorator.java). 
So far I managed to see sendMessage span in the traces in Tempo when I use sendMessage to my a2a application which is nice.
However, if I had my own custom span in an agent execution of my AgentExecutor, the whole trace is not consolidated with server span, it's 2 independent traces.
The main reason is that the server handling is executed inside a Vert.x router but the agent execution is done using a specific ExecutorService backed by A2A Internal executor.
Then, when the server handoff the execution to the agent executor, the opentelemetry context is not propagated and then is lost.
It results of having two different traces with their own spans and they are not linked at all.
- The server span in a trace (provided by a2a)
- My custom span in a trace

I would expect a single trace formed of two spans 

Opentelemetry context MUST be propagated between threads to ensure well formed traces with a valid parent span and the whole trace.
Since opentelemetry feature is not part of the server common modules, it's possible to define a SPI to implement for context propagation and implement it from the opentelemetry extra modules.

_No response_

### Additional context

_No response_

### Code of Conduct

vietk:
# Description

Introduce a context propagation mechanism when an agent execution is requested. 
The context propagation wraps the agent execution in a Runnable that can propagates a 'Context'

*  An OpenTelemetry implementation of `AgentExecutionContextPropagator` is provided in the opentelemetry module and is responsible to propagate the OpenTelemetry context between threads server thread and agent execution thread. This is needed to ensure well formed traces with a valid parent span.
*  The default implementation of `AgentExecutionContextPropagator` does not propagate any context and can be used when the opentelemetry feature is not enabled.
* Any further implementation could carry more than the OpenTelemetry context, e.g. MDC context, security context, etc.

Fixes #698

### DIFF
--- a/extras/opentelemetry/server/src/main/java/io/a2a/extras/opentelemetry/OtelAgentExecutionContextPropagator.java
+++ b/extras/opentelemetry/server/src/main/java/io/a2a/extras/opentelemetry/OtelAgentExecutionContextPropagator.java
@@ -1,0 +1,25 @@
+package io.a2a.extras.opentelemetry;
+
+import io.a2a.server.spi.AgentExecutionContextPropagator;
+import io.opentelemetry.context.Context;
+import jakarta.annotation.Priority;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Alternative;
+
+/**
+ * An implementation of AgentExecutionContextPropagator that uses OpenTelemetry's Context to propagate the execution
+ * context. This allows for proper context propagation across threads and asynchronous operations when using
+ * OpenTelemetry for tracing. The implementation simply delegates the wrapping OpenTelemetry's context propagation
+ * mechanism.
+ *
+ * @see Context#wrap(Runnable)
+ */
+@ApplicationScoped
+@Alternative
+@Priority(1)
+public class OtelAgentExecutionContextPropagator implements AgentExecutionContextPropagator {
+  @Override
+  public Runnable wrap(Runnable runnable) {
+    return Context.current().wrap(runnable);
+  }
+}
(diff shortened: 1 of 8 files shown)


extras/opentelemetry/server/src/main/java/io/a2a/extras/opentelemetry/OtelAgentExecutionContextPropagator.java
```diff
@@ -0,0 +1,25 @@
...
+ * context. This allows for proper context propagation across threads and asynchronous operations when using
+ * OpenTelemetry for tracing. The implementation simply delegates the wrapping OpenTelemetry's context propagation
+ * mechanism.
+ *
+ * @see Context#wrap(Runnable)
+ */
+@ApplicationScoped
+@Alternative
+@Priority(1)
+public class OtelAgentExecutionContextPropagator implements AgentExecutionContextPropagator {
```
salaboy, line 20: Do we really need this? I think we can always wrap the context if you add the opentelemetry API dependency to the project, no?

vietk: yep but because opentelemetry is provided as an extra dependency I was forced to use a SPI to implement it externally.
Adopting opentelemetry directly in the server is more a refactoring of the project that might not be suitable for a contribution

ehsavoie:
I think (but I need to finish some other stuff to check) that this is overkill and the best bet is to produce a ManagedExecutorService instead of the default @Internal ExecutorService

ehsavoie:
Superseded by https://github.com/a2aproject/a2a-java/pull/707 
@vietk @salaboy if you want to take a look and check that it correctly fixes your issue

vietk:
Nice, it's working fine on my side, I will close my PR

PR #700, opened 2026-03-04
fix!: rewrite PartTypeAdapter to use flat JSON format for Part serialization

### LINKED ISSUE #689: [Bug]: RestHandler#sendMessage returns part with both "text" and "filename" fields

ehsavoie:
Replace the nested "file" wrapper with flat fields (raw, url, filename, mediaType) aligned with the proto Part message schema and A2A spec.

Fixes #689 🦕

### DIFF
--- a/spec-grpc/src/test/java/io/a2a/grpc/utils/PartTypeAdapterTest.java
+++ b/spec-grpc/src/test/java/io/a2a/grpc/utils/PartTypeAdapterTest.java
@@ -1,0 +1,286 @@
+package io.a2a.grpc.utils;
+
+import io.a2a.jsonrpc.common.json.JsonProcessingException;
+import io.a2a.jsonrpc.common.json.JsonUtil;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import com.google.protobuf.ByteString;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Value;
+import com.google.protobuf.util.JsonFormat;
+import com.google.protobuf.util.Structs;
+import io.a2a.grpc.mapper.A2ACommonFieldMapper;
+import java.util.List;
+import java.util.Map;
+
+import io.a2a.spec.DataPart;
+import io.a2a.spec.FilePart;
+import io.a2a.spec.FileWithBytes;
+import io.a2a.spec.FileWithUri;
+import io.a2a.spec.Part;
+import io.a2a.spec.TextPart;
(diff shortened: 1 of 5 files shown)


spec-grpc/src/test/java/io/a2a/grpc/utils/PartTypeAdapterTest.java
```diff
@@ -0,0 +1,289 @@
...
+        String json = JsonFormat.printer().alwaysPrintFieldsWithNoPresence().omittingInsignificantWhitespace().print(builder);
+        Part<?> part = JsonUtil.fromJson(json, Part.class);
+        assertInstanceOf(FilePart.class, part);
+        FilePart filePart = (FilePart) part;
+        assertInstanceOf(FileWithBytes.class, filePart.file());
+        FileWithBytes fileWithBytes = (FileWithBytes) filePart.file();
+        assertEquals("image/png", fileWithBytes.mimeType());
+        assertEquals("diagram.png", fileWithBytes.name());
+        // Proto's JsonFormat encodes bytes as canonical base64 (zero trailing-padding bits),
+        // so "abc123==" (non-canonical) becomes "abc12w==" after the proto roundtrip.
```
jmesnil: what's that comment about? The builder passes `abc12w==` and I don't see `abc123==` in the test?

jmesnil (commented):
When I look at the test, I wonder if we should improve the `FileWithBytes` class to take in parameter a `java.io.File` and handles the Base64 conversion from the file's bytes internally?
That'd make for a much simpler UX. wdyt?

spec-grpc/src/test/java/io/a2a/grpc/utils/PartTypeAdapterTest.java
ehsavoie: yes, the issue was that abc123== was some 'wrong' base64 text that was translated to abc12w== due to padding

ehsavoie: I changed the content but forgot to remove the comment

ehsavoie:
I've created https://github.com/a2aproject/a2a-java/issues/702 to track your suggestion

PR #707, opened 2026-03-05
feat: propagate OTel context in async ops and fix TransportProtocol config injection

### LINKED ISSUE #698: [Feat]: Propagate open telemetry context to agent execution for consolidated tracing

ehsavoie:
- Replace direct TransportProtocol enum injection in AgentCardProducer with String injection + TransportProtocol.fromString() to avoid SmallRye Config's HyphenateEnumConverter failing on "HTTP+JSON" values passed via -D flags
- Add AsyncManagedExecutorProducer to the OpenTelemetry extras module, using MicroProfile ManagedExecutor to propagate OTel trace context across async boundaries (priority 20, overrides default AsyncExecutorProducer)
- Add microprofile-context-propagation-api dependency to opentelemetry/server

Fixes #698 🦕

### DIFF
--- a/examples/helloworld/server/src/main/java/io/a2a/examples/helloworld/AgentCardProducer.java
+++ b/examples/helloworld/server/src/main/java/io/a2a/examples/helloworld/AgentCardProducer.java
@@ -17,7 +17,7 @@ import java.util.List;
 public class AgentCardProducer {
 
     @ConfigProperty(name = "quarkus.agentcard.protocol", defaultValue = "JSONRPC")
-    TransportProtocol protocol;
+    String protocolStr;
 
     @Produces
     @PublicAgentCard
@@ -48,6 +48,7 @@ public class AgentCardProducer {
     }
 
     private AgentInterface getAgentInterface() {
+        TransportProtocol protocol = TransportProtocol.fromString(protocolStr);
         String url = switch (protocol) {
             case GRPC -> "localhost:9000";
             case JSONRPC, HTTP_JSON -> "http://localhost:9999";
--- a/extras/opentelemetry/server/src/main/java/io/a2a/extras/opentelemetry/AsyncManagedExecutorProducer.java
+++ b/extras/opentelemetry/server/src/main/java/io/a2a/extras/opentelemetry/AsyncManagedExecutorProducer.java
@@ -1,0 +1,65 @@
+package io.a2a.extras.opentelemetry;
+
+import io.a2a.server.util.async.Internal;
+import jakarta.annotation.PostConstruct;
+import jakarta.annotation.Priority;
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Alternative;
+import jakarta.enterprise.inject.Produces;
+import jakarta.inject.Inject;
+import java.util.concurrent.Executor;
+import org.eclipse.microprofile.context.ManagedExecutor;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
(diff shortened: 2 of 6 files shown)


salaboy:
@ehsavoie amazing stuff!

salaboy:
@ehsavoie I've tested this and it seems to be working.. it would be amazing if this can be included in the next release.

extras/opentelemetry/server/src/main/java/io/a2a/extras/opentelemetry/AsyncManagedExecutorProducer.java
```diff
@@ -0,0 +1,65 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(AsyncManagedExecutorProducer.class);
+    
+    @Inject
+    ManagedExecutor managedExecutor;
+
+    @PostConstruct
+    public void init() {
+        LOGGER.info("Initializing OpenTelemetry-aware ManagedExecutor for async operations");
+        if (managedExecutor == null) {
+            LOGGER.warn("ManagedExecutor not available - context propagation may not work correctly");
```
jmesnil, line 51: Is there a case where the `managerExecutor` would not be injected when `init` is called and would be injected before `produce` is called?

ehsavoie: If the ManagedExecutor can't be injected.

examples/helloworld/server/src/main/java/io/a2a/examples/helloworld/AgentCardProducer.java
```diff
@@ -17,7 +17,7 @@
 public class AgentCardProducer {
 
     @ConfigProperty(name = "quarkus.agentcard.protocol", defaultValue = "JSONRPC")
-    TransportProtocol protocol;
+    String protocolStr;
```
kabir, line 20: We could use a Converter https://github.com/microprofile/microprofile-config/blob/main/spec/src/main/asciidoc/converters.asciidoc but perhaps that is more trouble than it is worth

kabir (approved):

examples/helloworld/server/src/main/java/io/a2a/examples/helloworld/AgentCardProducer.java
ehsavoie: Too much trouble imho

PR #739, opened 2026-03-12
test: add unit tests for DeviceCodeOAuthFlow

### LINKED ISSUE #607: [Bug]: Missing tests for DeviceCodeOAuthFlow
### What happened?

In #606, `deviceCode` OAuth flow was added but there is no tests for it.

### Relevant log output

```shell

```

### Code of Conduct

neo1027144-creator:
Add unit tests for DeviceCodeOAuthFlow record class.

Test cases:
- Construction with all fields populated
- Null refreshUrl (optional field)
- Empty scopes map
- Null validation for required fields (deviceAuthorizationUrl, tokenUrl, scopes)
- Record equality and hashCode consistency

Follows the same testing patterns as FileWithBytesTest.

Fixes #607

### DIFF
--- a/spec/src/main/java/org/a2aproject/sdk/spec/DeviceCodeOAuthFlow.java
+++ b/spec/src/main/java/org/a2aproject/sdk/spec/DeviceCodeOAuthFlow.java
@@ -33,5 +33,16 @@ public record DeviceCodeOAuthFlow(String deviceAuthorizationUrl, String tokenUrl
         Assert.checkNotNullParam("deviceAuthorizationUrl", deviceAuthorizationUrl);
         Assert.checkNotNullParam("tokenUrl", tokenUrl);
         Assert.checkNotNullParam("scopes", scopes);
+        scopes = Map.copyOf(scopes);
+    }
+
+    /**
+     * Returns an unmodifiable copy of the scopes map.
+     *
+     * @return unmodifiable map of available OAuth scopes to their descriptions
+     */
+    @Override
+    public Map<String, String> scopes() {
+        return Map.copyOf(scopes);
     }
 }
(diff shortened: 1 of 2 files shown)


ehsavoie:
@neo1027144-creator can you please fix the gemini code review

neo1027144-creator:
Sorry for the delayed response — I've just addressed the review feedback. 
Thanks for the reminder!

@ehsavoie

spec/src/main/java/io/a2a/spec/DeviceCodeOAuthFlow.java
```diff
@@ -33,5 +33,6 @@ public record DeviceCodeOAuthFlow(String deviceAuthorizationUrl, String tokenUrl
         Assert.checkNotNullParam("deviceAuthorizationUrl", deviceAuthorizationUrl);
         Assert.checkNotNullParam("tokenUrl", tokenUrl);
         Assert.checkNotNullParam("scopes", scopes);
+        scopes = Map.copyOf(scopes);
```
ehsavoie, line 36: I don't think this 'work' but public Map<String, String> scopes() should be overriden to return Map.copyOf(scopes)

ehsavoie:
@neo1027144-creator Would you mind rebasing and fixing my comment.

neo1027144-creator:
Hi @ehsavoie, I've rebased onto the latest main and addressed your review comment. I've overridden `scopes()` to return `Map.copyOf(scopes)` as you suggested. I also kept the `Map.copyOf(scopes)` in the compact constructor to ensure the record's internal field is disconnected from the original mutable map reference (without this, external modifications to the original map would still be visible through the overridden accessor). Thanks for the review!

ehsavoie:
Thanks for the rebase, I'll take a look tomorrow :)
I think that the compact constructor doesn't work with what you did as the field is already set before the compact constructor code is called.

ehsavoie (approved):

PR #742, opened 2026-03-13
fix: synchronize ClientTaskManager to prevent race condition in concu…

kabir:
…rrent SSE event processing

ClientTaskManager was not thread-safe, causing intermittent test failures when
SSE events were delivered concurrently by the HTTP client. The race condition
occurred when multiple events (startWork, addArtifact, complete) were emitted
rapidly in streaming mode.

Without synchronization, concurrent threads could interleave read-modify-write
operations on currentTask, resulting in lost updates. Specifically, the artifact
update could be overwritten by a status update, leaving the final COMPLETED task
with no artifacts. The fast time to failure of the test 0.017s, indicates that the 
COMPLETED status update was received, and that the latch did not time out.

This manifested as intermittent failures in agent-to-agent tests where the test
received a COMPLETED task but extractTextFromTask() returned an empty string
because the artifact list was empty.

Fix: Added synchronized keyword to all ClientTaskManager methods (getCurrentTask,
saveTaskEvent variants, updateWithMessage) to ensure atomic processing of events
regardless of HTTP client threading behavior.

Impact: All client transports (JSON-RPC, gRPC, REST) in streaming mode
Validated: 100 consecutive test iterations passed on CI

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>


.github/workflows/test-race-condition.yml
```diff
@@ -0,0 +1,69 @@
+name: Test Race Condition Fix (Temporary)
```
jmesnil: @kabir do you plan to keep it in the CI or remove it before merging?

kabir: I thought I had removed it. Done now

PR #776, opened 2026-03-29
fix: preserve push notification event order per task (closes #775)

### LINKED ISSUE #775: [Bug]: pushnotification is not support order-preserving of update event
### What happened?

when status update or artifact update is sent through push notification, the request agent may not receivee those event in the order it produce. here is the code:

at MainEventBusProcessor.java line 337-357,the notification is sent asynchronous, regardless of whether the sending succeeds or fails. it may cause the wrong order of the event queue.
    private void sendPushNotification(String taskId, StreamingEventKind event) {
        Runnable pushTask = () -> {
            try {
                if (event != null) {
                    LOGGER.debug("Sending push notification for task {}", taskId);
                    pushSender.sendNotification(event);
                } else {
                    LOGGER.debug("Skipping push notification - event is null for task {}", taskId);
                }
            } catch (Exception e) {
                LOGGER.error("Error sending push notification for task {}", taskId, e);
                // Don't rethrow - push notifications are best-effort
            }
        };

        // Use custom executor if set (for tests), otherwise use default ForkJoinPool (async)
        if (pushNotificationExecutor != null) {
            pushNotificationExecutor.execute(pushTask);
        } else {
            CompletableFuture.runAsync(pushTask);
        }
    }

### Relevant log output

```shell

```

### Code of Conduct

perhaps468:
## Summary

Push notifications for streaming events (Task, TaskStatusUpdateEvent, TaskArtifactUpdateEvent) could arrive out of order on the client side because the previous implementation used `CompletableFuture.runAsync()` — which submits to the shared `ForkJoinPool.common()` pool, where events from concurrent tasks compete for the same thread pool and can be interleaved unpredictably.

This fix replaces the shared-pool approach with **per-task bounded queues**, each with a dedicated consumer thread, guaranteeing FIFO ordering per task and complete isolation between tasks.

---

## Problem

In the original implementation (`MainEventBusProcessor.sendPushNotification()`):

```java
CompletableFuture.runAsync(pushTask); // ForkJoinPool.common() — shared, unordered
```

`ForkJoinPool.common()` has a fixed parallelism of `Runtime.getRuntime().availableProcessors()` (typically 8–16). When multiple concurrent tasks each produce multiple events, those events race for thread pool slots. The consumer can receive task A's Event 2 before Event 1, purely as a scheduling artifact.

---

### Analysis of the Original Approach

The original implementation using `CompletableFuture.runAsync()` against `ForkJoinPool.common()` had clear advantages worth acknowledging:

**Original pros:**
- Zero memory overhead — no per-task queues or threads
- Zero thread management overhead — fully managed by the JVM
- Trivial to implement — one line of code
- Excellent throughput under low concurrency — no context-switching cost

**Original cons:**
- **Ordering not guaranteed across concurrent tasks** — events from different tasks can interleave in unpredictable ways due to thread scheduling
- **Non-deterministic delivery** — the same code path can produce different event sequences under load
- **No backpressure** — events pile up in the shared pool's queue with no per-task isolation

**Is this optimization necessary?** For a protocol like A2A where clients consume streams of events (text chunks arriving in order, artifacts building up incrementally), out-of-order delivery is a correctness bug, not just a performance issue. A client expecting `working → partial-completed → completed` could see `partial-completed → working → completed`, breaking expectations of any state-machine logic in the client. This is particularly critical for:
- Streaming responses (text parts must arrive in order for meaningful output)
- Multi-artifact tasks (each artifact appends sequentially)
- Client-side state machines that transition based on event type

Given the correctness implications, the per-task isolation is worth the added complexity.

### Option: Single-threaded executor per task (chosen)
Each task gets its own `LinkedBlockingQueue` + dedicated daemon thread.

**Pros:**
- Strict FIFO per task — no interleaving possible
- Complete isolation between tasks — one slow task never blocks another
- Simple, bounded memory (max 50 events per task, then oldest dropped)
- Lazy cleanup — threads expire after 60 seconds of inactivity

**Cons:**
- One daemon thread per active task (acceptable: daemon threads don't block JVM exit; 60s expiry limits growth)
- Thread-per-task model is not ideal for extreme scale (thousands of concurrent tasks), but realistic workloads rarely reach that

---

## Solution

Each task gets a `TaskPushNotifier` (inner class of `MainEventBusProcessor`) containing:

- `LinkedBlockingQueue<StreamingEventKind>` with capacity **50**
- Dedicated daemon consumer thread running a `consumeLoop()`
- `ConcurrentHashMap<String, TaskPushNotifier> notifiers` for per-task lookups
- 60-second inactivity expiry: expired notifiers are lazily replaced on next access

```
producer (MainEventBusProcessor thread)
  │
  ▼ offer()
LinkedBlockingQueue(50) ──► consumeLoop()
  │ (poll(10ms))              │
  │                       pushSender.sendNotification()
  ▼
(drop oldest if full)
```

**Overflow behavior:** If the queue is full when a new event arrives, the oldest pending event is silently dropped to make room. This is a deliberate design choice — a queue that stays full forever would deadlock the producer.

---

### New tests (`MainEventBusProcessorPushNotificationOrderTest`)

| Test | What it verifies | Result |
|---|---|---|
| `testSameTaskEventsOrdered` | Events for the same task arrive in strict FIFO order | ✅ |
| `testDifferentTasksUnaffected` | Events for different tasks are isolated and don't block each other | ✅ |
| `testQueueOverflowDropsOldest` | Queue holds at most 50 events; oldest are dropped on overflow | ✅ |

### Regression tests (`MainEventBusProcessorExceptionTest`)

All 3 existing exception handling tests continue to pass.

---

## Files Changed

- `server-common/src/main/java/io/a2a/server/events/MainEventBusProcessor.java`
  - Added `ConcurrentHashMap<String, TaskPushNotifier> notifiers`
  - Added `TaskPushNotifier` inner class (bounded queue + consumer thread + expiry)
  - Modified `sendPushNotification()` to route via per-task notifier
  - Modified `@PreDestroy stop()` to clean up all notifier threads

- `server-common/src/test/java/io/a2a/server/events/MainEventBusProcessorPushNotificationOrderTest.java` *(new)*
  - 3 tests covering FIFO ordering, task isolation, and overflow behavior

### DIFF
--- a/server-common/src/main/java/io/a2a/server/events/MainEventBusProcessor.java
+++ b/server-common/src/main/java/io/a2a/server/events/MainEventBusProcessor.java
@@ -1,6 +1,9 @@
 package io.a2a.server.events;
 
-import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executor;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
 
 import jakarta.annotation.Nullable;
 import jakarta.annotation.PostConstruct;
@@ -18,8 +21,8 @@ import io.a2a.spec.A2AServerException;
 import io.a2a.spec.Event;
 import io.a2a.spec.InternalError;
 import io.a2a.spec.Message;
-import io.a2a.spec.Task;
 import io.a2a.spec.StreamingEventKind;
+import io.a2a.spec.Task;
 import io.a2a.spec.TaskArtifactUpdateEvent;
 import io.a2a.spec.TaskStatusUpdateEvent;
 import org.slf4j.Logger;
@@ -69,10 +72,20 @@ public class MainEventBusProcessor implements Runnable {
 
     /**
      * Optional executor for push notifications.
-     * If null, uses default ForkJoinPool (async).
+     * If null, uses per-task queues with dedicated consumer threads.
      * Tests can inject a synchronous executor to ensure deterministic ordering.
      */
-    private volatile @Nullable java.util.concurrent.Executor pushNotificationExecutor = null;
+    private volatile @Nullable Executor pushNotificationExecutor = null;
+
+    /**
+     * Per-task push notifiers for ordered, isolated push notification delivery.
+     * Key: taskId, Value: TaskPushNotifier for that task.
+     * Notifiers are lazily cleaned up after NOTIFIER_EXPIRY_MS of inactivity.
+     */
+    private final ConcurrentHashMap<String, TaskPushNotifier> notifiers = new ConcurrentHashMap<>();
+
+    private static final int PUSH_NOTIFICATION_QUEUE_CAPACITY = 50;
+    private static final long NOTIFIER_EXPIRY_MS = 60_000;
 
     private MainEventBus eventBus;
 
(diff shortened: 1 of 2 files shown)


kabir:
@perhaps468 Sorry about totally missing this! I think I might have been on holiday at the time you opened it. Would you be able to rebase and resolve the Gemini review comments?

kabir:
@perhaps468 Closing due to inactivity. Feel free to reopen once rebased, and review comments resolved

PR #784, opened 2026-04-08
fix: return UnsupportedOperationError when subscribing to terminal tasks

kabir:
Per A2A spec, attempting to subscribe to a task in a terminal state (completed, failed, canceled, rejected) must return UnsupportedOperationError.

The issue had two parts:

1. DefaultRequestHandler checked terminal state AFTER queue operations began, causing errors to be delivered via SSE stream instead of immediately.

2. For streaming endpoints (subscribeToTask), immediate errors must be wrapped in SSE format so the client's SSE parser can process them and deliver to the error handler callback. Previously they were sent as plain JSON/HTTP responses which the SSE parser couldn't handle.

Changes:
- Move terminal state validation BEFORE queue operations in DefaultRequestHandler
- Wrap all immediate errors in SSE format for JSON-RPC routing layer
- Wrap all immediate errors in SSE format for REST transport
- Add integration test to verify fix across all three transports

Applied to HTTP+JSON, JSON-RPC, and gRPC transports.

Fixes #767

### DIFF
--- a/transport/jsonrpc/src/main/java/io/a2a/transport/jsonrpc/handler/JSONRPCHandler.java
+++ b/transport/jsonrpc/src/main/java/io/a2a/transport/jsonrpc/handler/JSONRPCHandler.java
@@ -388,9 +388,12 @@ public class JSONRPCHandler {
             // We can't use the convertingProcessor convenience method since that propagates any errors as an error handled
             // via Subscriber.onError() rather than as part of the SendStreamingResponse payload
             return convertToSendStreamingMessageResponse(request.getId(), publisher);
-        } catch (A2AError e) {
-            // Let routing layer catch TaskNotFoundError and UnsupportedOperationError to wrap in SSE format
+        } catch (TaskNotFoundError | UnsupportedOperationError e) {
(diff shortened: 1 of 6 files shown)


transport/jsonrpc/src/main/java/io/a2a/transport/jsonrpc/handler/JSONRPCHandler.java
```diff
@@ -389,7 +389,8 @@ public Flow.Publisher<SendStreamingMessageResponse> onSubscribeToTask(
             // via Subscriber.onError() rather than as part of the SendStreamingResponse payload
             return convertToSendStreamingMessageResponse(request.getId(), publisher);
         } catch (A2AError e) {
-            return ZeroPublisher.fromItems(new SendStreamingMessageResponse(request.getId(), e));
+            // Let routing layer catch TaskNotFoundError and UnsupportedOperationError to wrap in SSE format
+            throw e;
```
ehsavoie, line 393: Maybe this is a bit agressive if there is an internal error or something else

ehsavoie (approved):

ehsavoie:
Thanks @kabir

Issue #762, opened 2026-03-19
[Feat]: A2A v0.3.0 and v1.0.0 Coexistence


sduskis:
### Is your feature request related to a problem? Please describe.

My ecosystem has a bunch of agents talking A2A, primarily 0.3.0.  I'd like to evolve the ecosystem to export services for 0.3.0 and 1.0.0.  Is there a plan to support both in the same library?

### Describe the solution you'd like

It would be useful to have APIs that convert between 0.3.0 and 1.0.0.  I believe that most systems would start exposing 1.0.0 as a proxy for 0.3.0 and then migrate their implementation to be 1.0.0 compliant with a proxy that serves 0.3.0.  

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

kabir:
Hi @sduskis thank you for flagging this. We made a conscious decision in the run-up to 1.0 to not support 0.3.0 of the spec. 
However, we realise now that this might have been the wrong decision. Also, as the spec evolves beyond 1.0 we will need to support multiple protocols anyway. So we will look into this, and use this issue to discuss between ourselves.
It might not make it in time for our 1.0 release, but we think we can include it in a later 1.0.x release.

cammckenzie-atlassian:
hey @kabir ,
Have there been any more thoughts about this? We are in the process of working out how we will support the migration from 0.3 to 1.0.

Also, are there any ETAs on when there will be a stable 1.0 release?
Cheers

kabir:
Hi, we just had a long meeting about this this morning, and will work on this next :-)
I'm currently looking at https://github.com/a2aproject/a2a-java/issues/757 which is a big breaking change, but think we can release a beta once this works.
Then we have a few TCK issues to sort out, but hopefully the 1.0 final release isn't too far away

kabir:
@cammckenzie-atlassian @sduskis FYI we have started working on this, and the WIP is in https://github.com/a2aproject/a2a-java/pulls.
We've almost got the 0.3 TCK working against a server provisioned with the 0.3 protocol. 
Next step will be to look into what we need to do to be able to use both protocols at the same time.

cammckenzie-atlassian:
Amazing, thanks @kabir

kabir:
@sduskis @cammckenzie-atlassian The bulk of it is done in #805. Note there are a few followup issues in #854, #855, #856 and #857

cammckenzie-atlassian:
Great, thanks for the updates @kabir

[...]

PR #785, opened 2026-04-09
docs: PRD for A2A protocol 0.3 backward compatibility layer

jmesnil:
This relates to #762

### DIFF
--- a/.claude/architecture/compatibility_0.3.md
+++ b/.claude/architecture/compatibility_0.3.md
@@ -103,6 +103,23 @@ Users must explicitly check the `protocolVersion` field from the agent card and 
 
 `TASK_STATE_REJECTED` (v1.0-only) is mapped to `TASK_STATE_FAILED` when converting to v0.3 wire format. The original state is preserved in metadata (`"original_state": "REJECTED"`) so information is not entirely lost. Both are terminal states, so v0.3 clients can handle the result correctly.
 
+### Agent Card: v1.0 Only
+
+The agent card (`/.well-known/agent-card.json`) is produced only using the v1.0 format. The compat layer does not produce a v0.3-format agent card, nor does it serve a separate v0.3 agent card endpoint.
+
+A server that supports both versions advertises this via a single v1.0 agent card containing multiple `AgentInterface` entries — one per version — each with its own URL and `protocolVersion` field. v0.3 clients must be able to parse the v1.0 agent card format to discover their endpoint.
+
+**Pros:**
+- **Single source of truth**: one agent card at one well-known URL describes all supported versions and transports. No risk of agent cards drifting out of sync.
+- **Simpler server implementation**: no need for a separate v0.3 agent card endpoint, serializer, or CDI producer.
+- **Forward-looking**: encourages v0.3 clients to understand the v1.0 discovery format, which eases future migration.
+- **Reduced module scope**: the compat layer stays focused on runtime protocol translation, not discovery.
+
+**Cons:**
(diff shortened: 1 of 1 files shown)


.claude/architecture/compatibility_0.3.md
```diff
@@ -0,0 +1,561 @@
...
+
+---
+
+## Module Structure
+
+All compatibility code lives under a top-level `compat-0.3/` directory, mirroring the structure of the main SDK modules but containing **only** the code needed for 0.3 interop.
+
+```
+compat-0.3/
+├── pom.xml                          # Parent POM for all compat-0.3 submodules
```
kabir, line 135: We also will need the spec/ module

.claude/architecture/compatibility_0.3.md
```diff
@@ -0,0 +1,561 @@
...
+│   │   └── pom.xml
+│   └── rest/                        # Reference REST server
+│       └── pom.xml
+└── tck/                             # v0.3 conformance tests
+    └── pom.xml
+```
+
+### Java Package Convention
+
+All compat-0.3 code uses the `io.a2a.compat03` package root to avoid classpath collisions with the v1.0 modules:
```
kabir, line 175: Package should now be org.a2aproject.sdk.compat03

.claude/architecture/compatibility_0.3.md
```diff
@@ -0,0 +1,561 @@
...
+</dependency>
+
+<!-- Add v0.3 compatibility -->
+<dependency>
+    <groupId>org.a2aproject.sdk</groupId>
+    <artifactId>a2a-java-sdk-compat-0.3-reference-jsonrpc</artifactId>
+</dependency>
+```
+
+The same pattern applies for gRPC and REST transports.
```
kabir, line 310: You are mentioning transports here, but the examples a bit above are for the reference/ modules (rather than transport/)

.claude/architecture/compatibility_0.3.md
```diff
@@ -0,0 +1,561 @@
...
+
+**2. Declare both protocol versions in the AgentCard with separate URLs:**
+
+Each protocol version should use a distinct URL to avoid any dispatch ambiguity. The recommended pattern is to use a `/v0.3` path prefix for compat endpoints:
```
ehsavoie, line 314: 0.3.0 requires /v1/* as the path prefix. Also 1.0 has an optional tenant which could be confused with this prefix

kabir (changes requested):

.claude/architecture/compatibility_0.3.md
jmesnil: @ehsavoie My understanding is that with 0.3.0, you append the `/v1` prefix to the agent interface so this would not be a problem: you'd end up with: `http://example.org/v0.3/v1/message:send` for the HTTP+JSON endpoint.

Likewise for tenancy in the 1.0 spec.

.claude/architecture/compatibility_0.3.md
jmesnil: I'll fix it

.claude/architecture/compatibility_0.3.md
ehsavoie: Well in the RI the routing is done on /v1/ so http://example.org/v0.3/v1/ wouldn't match (maybe if there is a virtual host). 
JSONRPC routing is on / for the RI.
Why the tenant was introduced then for 1.0 if we could have used any prefix ?

jmesnil: I'm not sure we are talking about the same thing. Nothing should prevent a remote agent to be hosted on *any* URL. It does not have to be at the root of a domain.
My agent interfaces could be `http://example.org/agents/echo`. Any resolution of the endpoints should be relative to this context (ie `POST http://example.org/agents/echo/v1/message:send`).
The quarkus reference can be limited but that's not mandated by the spec.

kabir:
Included in #805

PR #788, opened 2026-04-10
fix: return TaskNotFoundError when SendMessage references non-existent taskId

### LINKED ISSUE #766: [Bug]: SendMessage with unknown taskId creates a new task instead of returning TaskNotFoundError
## Summary

When `SendMessage` includes a `taskId` that does not correspond to an existing task, the SUT creates a new task using the provided `taskId` instead of returning `TaskNotFoundError`. This affects all three transports.

Additionally, the status message in the response contains empty `contextId` and `taskId` fields, which should either be omitted or populated.

## Requirement

- **ID:** CORE-MULTI-004, CORE-SEND-002 (parametrized test only)
- **Section:** 3.4.2 — Task Identifier Semantics
- **Level:** MUST
- **Spec:** https://github.com/a2aproject/A2A/blob/173695755607e884aa9acf8ce4feed90e32727a1/docs/specification.md#342-task-identifier-semantics

## Specification

> When a client includes a `taskId` in a Message, it **MUST** reference an existing task.
> Agents **MUST** return a `TaskNotFoundError` if the provided `taskId` does not correspond to an existing task.
> Client-provided `taskId` values for creating new tasks is **NOT** supported.

## Expected behavior

SendMessage with a non-existent `taskId` should return `TaskNotFoundError`:
- JSON-RPC: error code `-32001`
- gRPC: status `NOT_FOUND`
- HTTP+JSON: status `404 Not Found`

## Actual behavior

The SUT creates a new task using the client-provided `taskId`, processes it (reaching `TASK_STATE_COMPLETED`), and returns success. The status message in the response also contains empty `contextId` and `taskId` fields:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "task": {
      "id": "does-not-exist-99999",
      "contextId": "b37f0cf9-5daf-4e98-a9d8-c05e1cb54eea",
      "status": {
        "state": "TASK_STATE_COMPLETED",
        "message": {
          "messageId": "30714123-0f06-477a-a81f-239fb97e1b5b",
          "contextId": "",
          "taskId": "",
          "role": "ROLE_AGENT",
          "parts": [{"text": "Unhandled messageId prefix: test-repro-001"}],
          ...
        }
      }
    }
  }
}
```

On subsequent calls with the same `taskId`, it returns `UnsupportedOperationError` because the task is now terminal.

## Reproducer

```bash
# Send a message with a nonexistent taskId
# Expected: TaskNotFoundError (-32001)
# Actual: SUT creates a new task with the client-provided taskId and returns success
curl -s -X POST http://localhost:9999/ \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":"1","method":"SendMessage","params":{"message":{"messageId":"test-repro-001","taskId":"does-not-exist-99999","role":"ROLE_USER","parts":[{"text":"Hello"}]}}}'

# Response: 200 OK with task created using id "does-not-exist-99999"
#   - task.status.message.contextId is empty ("")
#   - task.status.message.taskId is empty ("")
# Expected: {"jsonrpc":"2.0","id":1,"error":{"code":-32001,"message":"Task not found"}}
```

## TCK test

`tests/compatibility/core_operations/test_requirements.py::test_must_requirement[CORE-MULTI-004-grpc]`
`tests/compatibility/core_operations/test_requirements.py::test_must_requirement[CORE-MULTI-004-jsonrpc]`
`tests/compatibility/core_operations/test_requirements.py::test_must_requirement[CORE-MULTI-004-http_json]`

### Relevant log output

```shell

```

### Code of Conduct

yyy9942:
Per A2A spec section 3.4.2, when a client includes a taskId in a Message, it MUST reference an existing task. The SDK was creating a new task with that id instead.

Added a guard in DefaultRequestHandler.initMessageSend() that throws TaskNotFoundError when the provided taskId does not reference an existing task.

This fixes the issue on all three transports (JSON-RPC, gRPC, HTTP+JSON) since they all use the same DefaultRequestHandler code path.

Also fixes a related contextId bug: when a follow-up message had a taskId but no contextId, RequestContext.Builder.build() was generating a fresh UUID, so the agent executor saw the wrong contextId. The rebuild now seeds contextId from the stored task.

Existing tests that relied on the old taskId behavior have been updated.

Fixes #766

### DIFF
--- a/server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/DefaultRequestHandler.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/DefaultRequestHandler.java
@@ -1021,56 +1021,26 @@ public class DefaultRequestHandler implements RequestHandler {
     }
 
     private MessageSendSetup initMessageSend(MessageSendParams params, ServerCallContext context) throws A2AError {
-        // Build RequestContext FIRST to get the real taskId (auto-generated if not provided)
-        // This eliminates the need for temporary IDs - we use the same UUID throughout
+        Task task = validateRequestedTask(params);
+        MessageSendParams requestParams = task == null ? params : normalizeRequestParamsForTask(params, task);
+
         RequestContext requestContext = requestContextBuilder.get()
-                .setParams(params)
-                .setTaskId(params.message().taskId())  // Use client's ID or let RequestContext generate
-                .setContextId(params.message().contextId())
-                .setTask(null)  // Will be set below after TaskManager retrieves it
+                .setParams(requestParams)
+                .setTaskId(requestParams.message().taskId())
+                .setContextId(task != null ? task.contextId() : requestParams.message().contextId())
+                .setTask(task)
                 .setServerCallContext(context)
                 .build();
 
-        // Get the actual taskId from RequestContext (either from client or auto-generated)
-        // RequestContext.build() guarantees taskId is non-null via checkOrGenerateTaskId()
-        String taskId = java.util.Objects.requireNonNull(
-                requestContext.getTaskId(), "TaskId must be non-null after RequestContext.build()");
+        String taskId = Objects.requireNonNull(requestContext.getTaskId());
 
-        // Create TaskManager with the real taskId
         TaskManager taskManager = new TaskManager(
                 taskId,
-                params.message().contextId(),
+                requestContext.getContextId(),
                 taskStore,
-                params.message());
+                requestParams.message());
 
-        Task task = taskManager.getTask();
         if (task != null) {
-            // Reject messages to tasks that are in a terminal state (completed, canceled, rejected, failed).
-            // Per A2A spec section 3.1.1 (CORE-SEND-002): the SDK MUST return UnsupportedOperationError
-            // before forwarding the message to the AgentExecutor.
-            if (task.status().state().isFinal()) {
-                throw new UnsupportedOperationError(
-                        null,
-                        "Cannot send message to task " + task.id() +
-                        " - task is in a terminal state: " + task.status().state(),
-                        null);
-            }
-
-            // Validate contextId matches the existing task's contextId
-            String messageContextId = params.message().contextId();
-            if (messageContextId != null && !messageContextId.equals(task.contextId())) {
-                throw new InvalidParamsError(String.format(
-                        "Message has a mismatched context ID (Task %s has contextId %s but message has contextId %s)",
-                        task.id(), task.contextId(), messageContextId));
-            }
-
-            // Per spec CORE-SEND-002: Reject messages to tasks in terminal states
-            if (task.status().state().isFinal()) {
-                throw new UnsupportedOperationError(null, String.format(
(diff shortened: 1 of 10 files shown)


yyy9942:
This may have introduced, or at least exposed, a transport-specific regression in the new test path.

In GitHub Actions run 24245000423 on April 13, 2026, the failing test is: `QuarkusA2AJSONRPCVertxTest.testAgentToAgentLocalHandling`

Failure: `Local handling should complete within timeout ==> expected: <true> but was: <false>`

What makes this look PR-related:
- `QuarkusA2AJSONRPCJdkTest` passes in the same run
- #788 changed this test flow from client-provided `taskId/contextId` and taskId-based routing to server-generated task IDs and message-content routing (`a2a-local:`)
- the failure is a timeout, which suggests the Vert.x JSON-RPC path may not be delivering the final task event/latch signal in this new flow

Minimal reproduction candidate:
```bash
  mvn -pl reference/jsonrpc -am \
    -Dtest=QuarkusA2AJSONRPCVertxTest#testAgentToAgentLocalHandling \
    -Dsurefire.failIfNoSpecifiedTests=false \
    test
```

The later MultiInstanceReplicationTest failure looks secondary:
- its Docker build fails because target/quarkus-app/lib/ is missing
- multiple upstream modules were already banned/skipped due to the earlier failure

I haven't reproduced it locally yet, but the CI failure looks pretty tightly aligned with the test-path changes in this PR.

server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/DefaultRequestHandler.java
```diff
@@ -1044,6 +1044,14 @@ private MessageSendSetup initMessageSend(MessageSendParams params, ServerCallCon
                 params.message());
 
         Task task = taskManager.getTask();
+
+        // Per A2A spec section 3.4.2: when a client includes a taskId in a
+        // Message, it MUST reference an existing task. Client-provided taskId
+        // for creating new tasks is not supported.
+        if (task == null && params.message().taskId() != null) {
```
jmesnil: I find this check a bit convoluted.
It can be simplified by verifying at the start of the method that if `params.message().taskId()` is not null, there is a task returned by `taskStore.get(params.message().taskId())`.
After that, the request context can be safely created from either a correct message's taskId or an autogenerated taskId.

yyy9942: @jmesnil Simplified the check as suggested. Extracted the taskId validation into `validateRequestedTask()` at the top of the method - if `params.message().taskId()` is non-null, it verifies the task exists, validates contextId consistency, and rejects terminal-state tasks upfront. After that, `initMessageSend()` can safely work with either a validated existing task or proceed with auto-generated identifiers for new tasks.

Also pulled out `normalizeRequestParamsForTask()` to align message params with the task's canonical IDs before building the RequestContext, and merged the two consecutive `if (task != null)` blocks into one.

ehsavoie (approved):

ehsavoie:
Thanks @yyy9942

PR #801, opened 2026-04-17
feat: preserve initial message in task history by reusing TaskManager

kabir:
Pass TaskManager from DefaultRequestHandler to MainEventBusProcessor via EventQueue, allowing the initial message to be included in task history when the agent submits the task.

Clear TaskManager from MainQueue on ChildQueue close to prevent memory leak, as MainQueue can live for a long time but TaskManager is only needed for the first event.

### DIFF
--- a/server-common/src/main/java/org/a2aproject/sdk/server/events/EventQueue.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/events/EventQueue.java
@@ -10,6 +10,7 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.a2aproject.sdk.server.tasks.TaskManager;
 import org.a2aproject.sdk.server.tasks.TaskStateProvider;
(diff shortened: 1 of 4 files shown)


kabir:
@jmesnil pointed out:

> According to https://a2a-protocol.org/latest/specification/#37-messages-and-artifacts,
The Task History field contains Messages exchanged during task execution. However, not all Messages are guaranteed to be persisted in the Task history; for example, transient informational messages may not be stored. Messages exchanged prior to task creation may not be stored in Task history. The agent is responsible to determine which Messages are persisted in the Task History.

This PR was driven by the work on the 0.3 compatibility layer. Instead what I have done there is to have the agent executor include the message when calling AgentEmitter.submit()

kabir:
@jmesnil pointed out:

> According to https://a2a-protocol.org/latest/specification/#37-messages-and-artifacts,
The Task History field contains Messages exchanged during task execution. However, not all Messages are guaranteed to be persisted in the Task history; for example, transient informational messages may not be stored. Messages exchanged prior to task creation may not be stored in Task history. The agent is responsible to determine which Messages are persisted in the Task History.

So

PR #806, opened 2026-04-21
feat(http-client): add Android support and implement SPI for A2AHttpClient

### LINKED ISSUE #793: [Feat]: Add Android support for A2A HTTP Client and decouple JdkA2AHttpClient (v0.3.x)
## Problem

Currently, the A2A Java SDK hardcodes the use of `JdkA2AHttpClient` in several places (such as `A2A.java` and various transport providers). This client relies on `java.net.http.HttpClient`, which is not available in standard Android environments, making the SDK unusable on Android. 
Additionally, when running on Android, Jackson fails to deserialize Java records properly without explicit property annotations, leading to parsing errors for spec models.

## Proposed Solution

- **SPI Pattern**: Introduce `A2AHttpClientFactory` and `A2AHttpClientProvider` using Java's `ServiceLoader` (SPI) to dynamically load the best HTTP client available on the classpath.
- **Android Client**: Implement `AndroidA2AHttpClient` using `HttpURLConnection` (which is safe for Android).
- **Jackson Fixes**: Add `@JsonProperty` annotations to records in the `spec/` module to ensure they can be parsed on Android.

_No response_

## Goal

The goal is to make the `http-client` module compatible with Android by:
1. Decoupling the HTTP client implementation so it is not hardcoded to the JDK client.
2. Providing an Android-compatible HTTP client implementation.
3. Fixing the Jackson parsing issues for records on Android.

### Code of Conduct

sherryfox:
# Summary of Changes
This PR introduces Android support for the A2A HTTP client, decouples the hardcoded JDK client using the Service Provider Interface (SPI) pattern, and aligns with the upcoming Gson migration by rebasing onto the `0.3-workflows` branch.
## 🌟 Key Highlights
- **Android Support**: Added a dedicated HTTP client for Android using `HttpURLConnection`.
- **SPI Pattern**: Decoupled `JdkA2AHttpClient` and introduced a factory to resolve the best client automatically.
- **Gson Alignment**: Removed Jackson annotations from spec records to match the Gson fork approach.
---
### 1. HTTP Client Factory & SPI Pattern
Introduced a factory and provider interface to allow pluggable HTTP client implementations via `ServiceLoader`.
*   **`A2AHttpClientFactory.java`**: Created to load implementations of `A2AHttpClientProvider` and select the one with the highest priority.
*   **`A2AHttpClientProvider.java`**: Introduced as the interface for providers.
*   **Service Configurations**: Added files to register the providers in:
    *   `http-client/src/main/resources/META-INF/services/io.a2a.client.http.A2AHttpClientProvider`
    *   `http-client/src/main/resources_android/META-INF/services/io.a2a.client.http.A2AHttpClientProvider`
### 2. Decoupling `JdkA2AHttpClient`
Removed hardcoded dependencies on `JdkA2AHttpClient` across the codebase.
*   `JdkA2AHttpClient` now implements `A2AHttpClientProvider` with a default priority of `0`.
*   Updated the following files to use `A2AHttpClientFactory.create()` instead of direct instantiation:
    *   `A2A.java` (in `client/base`)
    *   `A2ACardResolver.java` (in `http-client`)
    *   `JSONRPCTransportProvider.java` (in `client/transport/jsonrpc`)
    *   `RestTransportProvider.java` (in `client/transport/rest`)
### 3. `AndroidA2AHttpClient`
*   Introduced `AndroidA2AHttpClient.java` using `HttpURLConnection` (which does not have forbidden dependencies on Android).
*   It implements `A2AHttpClientProvider` with a priority of `100`, making it the preferred choice in Android environments.
### 4. Alignment with Gson Migration
*   Removed Jackson annotations from `spec` records (like `AgentCapabilities`, `TaskStatus`, etc.) as they are not needed in the Gson environment.
### 5. Documentation & Fixes
*   **Documentation**: Updated `README.md` to reflect that the default HTTP client is resolved automatically via the factory rather than being hardcoded.
*   **Kafka Version Update**: Kept the Kafka version at `4.2.0` (from the fork) in `examples/cloud-deployment/k8s/02-kafka.yaml` to ensure compatibility with the CI environment.
---
## 🧪 Verification Results
*   **Build**: Succeeded with `mvn clean install -DskipTests=true`.
*   **Tests**: All tests in the `http-client` module passed successfully (`mvn test -pl http-client`), including a new test verifying the factory behavior.

# Description

- [x] Follow the [`CONTRIBUTING` Guide](../CONTRIBUTING.md).
- [x] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [x] Ensure the tests pass

Fixes #793 🦕


client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransportProvider.java
```diff
@@ -17,7 +17,7 @@ public String getTransportProtocol() {
     public RestTransport create(RestTransportConfig clientTransportConfig, AgentCard agentCard, String agentUrl) throws A2AClientException {
         RestTransportConfig transportConfig = clientTransportConfig;
          if (transportConfig == null) {
-            transportConfig = new RestTransportConfig(new JdkA2AHttpClient());
+            transportConfig = new RestTransportConfig(A2AHttpClientFactory.create());
         }
         return new RestTransport(clientTransportConfig.getHttpClient(), agentCard, agentUrl, transportConfig.getInterceptors());
```
ehsavoie: This shoud use transportConfig not clientTransportConfig

http-client/src/main/java/io/a2a/client/http/AndroidA2AHttpClient.java
```diff
@@ -0,0 +1,268 @@
...
+
+    @Override
+    public CompletableFuture<Void> postAsyncSSE(
+        Consumer<String> messageConsumer,
+        Consumer<Throwable> errorConsumer,
+        Runnable completeRunnable)
+        throws IOException {
+      HttpURLConnection connection = createConnection("POST");
+      connection.setDoOutput(true);
+      try (OutputStream os = connection.getOutputStream()) {
```
ehsavoie: This will do a connect and read the body before executeAsyncSSE is called, but you are still configuring the connection in executeAsyncSSE

extras/http-client-android/src/main/java/io/a2a/client/http/AndroidA2AHttpClient.java
```diff
@@ -0,0 +1,268 @@
...
+      URL urlObj;
+      try {
+        urlObj = new URI(url).toURL();
+      } catch (URISyntaxException e) {
+        throw new MalformedURLException("Invalid URL: " + url);
+      }
+      HttpURLConnection connection = (HttpURLConnection) urlObj.openConnection();
+      connection.setRequestMethod(method);
+      connection.setConnectTimeout(15000); // 15 seconds
+      connection.setReadTimeout(60000);    // 60 seconds
```
ehsavoie, line 90: Is this sufficient for SSE ?

ehsavoie (changes requested):
Also I'm wondering about moving the android client to a specific module like we did for vertx in main. So maybe we could run tests against it in the same way ?

client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransportProvider.java
sherryfox: Good point!

http-client/src/main/java/io/a2a/client/http/AndroidA2AHttpClient.java
sherryfox: Refactored

extras/http-client-android/src/main/java/io/a2a/client/http/AndroidA2AHttpClient.java
sherryfox: Keeping in mind that 60s for streaming should apply to time between chunks, then technically it should be enough but it also depends on server implementation. For Android, to not block the threads and consume the resources, we can try to set the limit to 5 minutes (to avoid infinite timeout) - but then it might create user non-friendly experience. My proposal is by keeping the timeout relatively short (60s) and the application could resubscribe on this error, giving the app control over how many times they would want to retry, how long to wait between then and how to show that to a user. Wdyt?

sherryfox:
> Also I'm wondering about moving the android client to a specific module like we did for vertx in main. So maybe we could run tests against it in the same way ?

Moved to `extras/http-client-android` following the pattern from main

extras/http-client-android/src/main/java/io/a2a/client/http/AndroidA2AHttpClient.java
ehsavoie: Sorry, wasn't really thinking as an Android user. It makes sense as it is, it would be up to the 'client' to resubscribe

client/transport/jsonrpc/src/main/java/io/a2a/client/transport/jsonrpc/JSONRPCTransportConfigBuilder.java
```diff
@@ -1,6 +1,7 @@
 package io.a2a.client.transport.jsonrpc;
 
 import io.a2a.client.http.A2AHttpClient;
+import io.a2a.client.http.A2AHttpClientFactory;
 import io.a2a.client.http.JdkA2AHttpClient;
```
ehsavoie: this isn't used anymore

client/transport/rest/src/main/java/io/a2a/client/transport/rest/RestTransportConfigBuilder.java
```diff
@@ -1,6 +1,7 @@
 package io.a2a.client.transport.rest;
 
 import io.a2a.client.http.A2AHttpClient;
+import io.a2a.client.http.A2AHttpClientFactory;
 import io.a2a.client.http.JdkA2AHttpClient;
```
ehsavoie: This isn't used anymore

ehsavoie (changes requested):
2 samml imports to remove and we are good to go

ehsavoie (approved):

PR #837, opened 2026-04-29
fix: support complete agent card URL in A2ACardResolver

ehsavoie:
URI.resolve() silently drops intermediate path segments when the
resolved path starts with '/', causing incorrect URLs for agents
hosted under a path prefix (e.g. /spec03). Replace with direct
concatenation and add a pass-through guard when the supplied URL
already ends with the agent card path.

Fixes #771 🦕

### DIFF
--- a/http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
+++ b/http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
@@ -1,10 +1,11 @@
 package org.a2aproject.sdk.client.http;
 
-
 import java.io.IOException;
-import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.HashMap;
 import java.util.Map;
+
+import static org.a2aproject.sdk.util.Assert.checkNotNullParam;
 
 import org.a2aproject.sdk.grpc.utils.JSONRPCUtils;
 import org.a2aproject.sdk.grpc.utils.ProtoUtils;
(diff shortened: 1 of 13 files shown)


jmesnil (commented):
I'm not sure if we should make the card resolver more complex.
I see 2 use case for fetching an agent card:
* use the spec's well-known endpoint that is appended to whatever the baseUrl is (it can be http://example.com, http://example.com/agents/v1/foo)
* point to a specific URL that returns an agent card. This does not have to end with `.well-known/.agent-card.json`. This URL can be fetched from a registry for example. In that case, there is no URL resolution at all, it's used verbatim.

So the API would be:
* A2AResolver(String url) (and all its variant)
* AgentCard getWellKnownAgentCard() -> append `.well-known/.agent-card.json to the base URL
* AgentCard getAgentCard() -> use the URL verbatim

That's be a API break change so maybe we could have just a new method:

* AgentCard getAgentCard(boolean fromWellKnownLocation)
  * getAgentCard() delegates to getAgentCard(true)
  * the resolution of the URL is deferred until getAgentCard(boolean) is called to figure out which URL to connect to.

ehsavoie:
@jmesnil I broke the API because I think it is cleaner and clearer that way

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
```diff
@@ -1,184 +1,316 @@
...
+     */
+    public AgentCard getConfiguredAgentCard() throws A2AClientError, A2AClientJSONError {
+        return fetchAgentCard(this.configuredAgentCardUrl);
+    }
+
+    /**
+     * @deprecated Use {@link #getConfiguredAgentCard()} for fetching the configured card URL, or
+     * {@link #getWellKnownAgentCard()} for the standard well-known endpoint.
      */
+    @Deprecated
```
kabir: Could we just not keep this method and remove the two new methods?
It seems a bit cumbersome to have the user choose between getConfiguredAgentCard() and getWellKnownAgentCard().
In my mind, if one is configured we return that, and if not we fallback to the well known one

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
```diff
@@ -1,184 +1,316 @@
...
+
+        /**
+         * Sets the authentication headers to use when fetching the agent card.
+         *
+         * @param authHeaders the authentication headers, may be null
+         * @return this builder
+         */
+        public Builder authHeaders(@Nullable Map<String, String> authHeaders) {
+            if (authHeaders != null) {
+                this.authHeaders = new java.util.HashMap<>(authHeaders);
```
kabir: import HashMap

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
```diff
@@ -1,184 +1,316 @@
...
+        /**
+         * Adds a single authentication header.
+         *
+         * @param name the header name
+         * @param value the header value
+         * @return this builder
+         */
+        public Builder authHeader(String name, String value) {
+            if (this.authHeaders == null) {
+                this.authHeaders = new java.util.HashMap<>();
```
kabir: import HashMap

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
```diff
@@ -1,184 +1,316 @@
...
+     * {@link #getWellKnownAgentCard()} for the standard well-known endpoint.
      */
+    @Deprecated
     public AgentCard getAgentCard() throws A2AClientError, A2AClientJSONError {
+        return getConfiguredAgentCard();
+    }
+
+    private static URI resolveAgentCardUrl(URI baseUri, String agentCardPath) throws URISyntaxException {
+        String normalizedAgentCardPath = agentCardPath.startsWith("/") ? agentCardPath : "/" + agentCardPath;
+        if (baseUri.getPath() != null && baseUri.getPath().endsWith(normalizedAgentCardPath)) {
```
kabir: Apparently:

> This fires on any path that ends with the agent card path, which can produce false positives. E.g., a base URL path of /not-well-known/agent-card.json would match /.well-known/agent-card.json via endsWith. More practically, if a user passes a full well-known URL as baseUrl and provides a different custom agentCardPath, the guard wouldn't catch it and the well-known suffix would remain in the base, producing a broken concatenation like https://example.com/spec03/.well-known/agent-card.json/custom/card.json.

kabir (commented):
Also since I think this changes the API the prefix to the PR title should be "fix!" rather than "fix"

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
ehsavoie: We always have both since both are based on the builtBaseUri

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
ehsavoie: Should be fixed now

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
```diff
@@ -1,184 +1,327 @@
...
+            this.configuredAgentCardUrl = (agentCardPath == null || agentCardPath.isBlank())
+                    ? builtBaseUri.toString()
+                    : resolveAgentCardUrl(builtBaseUri, agentCardPath).toString();
+
+            LOGGER.debug("Initialized A2ACardResolver with wellKnownUrl={}, configuredAgentCardUrl={}", wellKnownUrl, configuredAgentCardUrl);
+        } catch (URISyntaxException e) {
+            throw new A2AClientError("Invalid agent URL", e);
+        }
+        this.authHeaders = authHeaders;
     }
```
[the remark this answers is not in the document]
ehsavoie: I disagree, we could have the final url passed as the base url without any agentPath

http-client/src/main/java/org/a2aproject/sdk/client/http/A2ACardResolver.java
```diff
@@ -1,184 +1,327 @@
...
+         *
+         * @param authHeaders the authentication headers, may be null
+         * @return this builder
+         */
+        public Builder authHeaders(@Nullable Map<String, String> authHeaders) {
+            if (authHeaders != null) {
+                this.authHeaders = new HashMap<>(authHeaders);
+            }
+            return this;
+        }
```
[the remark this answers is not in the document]
ehsavoie: We don't want null headers

PR #838, opened 2026-04-29
feat: Migrate Reactive Routes to Vert.X Web and add security tests

kabir:


reference/jsonrpc/src/main/java/org/a2aproject/sdk/server/apps/quarkus/A2AServerRoutes.java
```diff
@@ -186,6 +191,53 @@ public class A2AServerRoutes {
...
+            .handler(ctx -> {
+                try {
+                    String agentCard = getAgentCard(ctx);
+                    ctx.response()
+                        .setStatusCode(200)
+                        .putHeader(CONTENT_TYPE, APPLICATION_JSON)
+                        .end(agentCard);
+                } catch (JsonProcessingException e) {
+                    ctx.response().setStatusCode(500).end("Internal Server Error");
+                }
```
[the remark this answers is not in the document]
kabir: @gemini-code-assist With @Route, the method's return value (String) was automatically sent as the response body by the framework. If JsonProcessingException was  
  thrown, Quarkus Reactive Routes would have returned it as a 500 with the framework's default error handling — so the current 500 behavior is  actually consistent with the old behavior.                                                                                                    
                                                                                                                                              
  The JsonProcessingException here can only happen when serializing the AgentCard to JSON (not parsing incoming data), so it's genuinely an     
  internal server error — the server failed to serialize its own data. A 400 would be wrong here since the client didn't send anything invalid.
  Gemini's suggestion is incorrect in this context.

reference/common/src/main/java/org/a2aproject/sdk/server/common/quarkus/VertxSecurityHelper.java
```diff
@@ -0,0 +1,160 @@
...
+     * but would block the event loop on event loop threads.
+     *
+     * @param ctx the Vert.x routing context containing the HTTP request
+     * @param task the code to execute within the authenticated request context
+     * @throws io.quarkus.security.UnauthorizedException if authentication fails
+     * @throws io.quarkus.security.ForbiddenException if authorization fails
+     * @throws RuntimeException if the task throws an exception
+     */
+    public void runInRequestContext(RoutingContext ctx, Runnable task) {
+        ManagedContext requestContext = Arc.container().requestContext();
```
ehsavoie, line 106: We should protect the event loop thread:

 if (Context.isOnEventLoopThread()) {
        throw new IllegalStateException("Cannot perform blocking authentication on event loop thread. Use blockingHandler().");
    }

reference/common/src/main/java/org/a2aproject/sdk/server/common/quarkus/VertxSecurityHelper.java
```diff
@@ -0,0 +1,160 @@
...
+    public static void handleAuthError(RoutingContext ctx, Exception e) {
+        if (!ctx.response().ended()) {
+            if (e instanceof io.quarkus.security.ForbiddenException) {
+                ctx.response()
+                    .setStatusCode(403)
+                    .end();
+            } else {
+                ctx.response()
+                    .setStatusCode(401)
+                    .putHeader("WWW-Authenticate", "Basic realm=\"Quarkus\"")
```
ehsavoie: Shouldn't the authentication realm be configurable ??

ehsavoie (changes requested):
TestIdentityProvider is duplicated in reference/jsonrpc and reference/rest. It belongs in tests/server-common alongside AbstractA2AServerWithAuthTest.


AuthTestProfile is duplicated across all three modules with near-identical content. Shared config should move to tests/server-common.

ehsavoie (approved):

PR #858, opened 2026-05-07
fix: guard against JsonNull when parsing response error field  

RainYuY:
# Description

#  Root Cause

  The serializer and deserializer are inconsistent:
  - A2AErrorTypeAdapter.write() emits "error": null for null errors
  - parseResponseBody() assumes the key's presence implies a non-null object

  The correct long-term fix is to also stop emitting "error": null in
  A2AErrorTypeAdapter.write(), but this client-side guard is sufficient to
  prevent the crash.

#  Impact

  Without this fix, every non-streaming A2A call fails with ClassCastException
  regardless of whether the response is actually an error.

--------------

- [ ] Follow the [`CONTRIBUTING` Guide](../CONTRIBUTING.md).
- [ ] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [ ] Ensure the tests pass
- [ ] Appropriate READMEs were updated (if necessary)


kabir:
Thanks @RainYuY !

PR #935, opened 2026-06-16
feat!: add TaskAuthorizationProvider SPI for per-user task authorization

kabir:
Add an opt-in SPI for per-user task authorization, enforced via a CDI decorator around RequestHandler.

- TaskOperation enum identifies which operation triggered the check
- TaskAuthorizationProvider interface defines checkRead/checkWrite/ checkCreate/recordOwnership/isTaskRecorded methods
- AuthorizationRequestHandlerDecorator (@Decorator @Priority(50)) wraps RequestHandler, calling the provider before/after each method
- For onListTasks, filtering is pushed to TaskStore.list() via a new ServerCallContext parameter
- InMemoryTaskStore filters all tasks before pagination (exact results)
- JpaDatabaseTaskStore uses an iterative fetch loop to accumulate pageSize authorized results across DB pages (correct page sizes)
- Streaming ownership recording wraps the publisher to call recordOwnership on the first task event
- QuarkusCallContextFactory propagates Quarkus SecurityIdentity into gRPC ServerCallContext, so authorization checks see the authenticated user. Uses Instance<SecurityIdentity> for optional injection in apps without security configured
- Integration tests cover all 3 transports (JSON-RPC, REST, gRPC) across reference, multiversion, and compat-0.3 modules

BREAKING CHANGE: TaskStore.list() now requires a ServerCallContext parameter. Custom TaskStore implementations must update their list() signature and apply authorization filtering when a TaskAuthorizationProvider is present.

### DIFF
--- a/compat-0.3/reference/grpc/src/main/java/org/a2aproject/sdk/compat03/server/grpc/quarkus/QuarkusCallContextFactory_v0_3.java
+++ b/compat-0.3/reference/grpc/src/main/java/org/a2aproject/sdk/compat03/server/grpc/quarkus/QuarkusCallContextFactory_v0_3.java
@@ -5,7 +5,6 @@ import static org.a2aproject.sdk.server.ServerCallContext.TRANSPORT_KEY;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
-import java.util.Set;
 
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.enterprise.inject.Instance;
@@ -28,7 +27,7 @@ import org.a2aproject.sdk.spec.TransportProtocol;
 public class QuarkusCallContextFactory_v0_3 implements CallContextFactory_v0_3 {
 
     @Inject
-    Instance<SecurityIdentity> securityIdentityInstance;
+    private Instance<SecurityIdentity> securityIdentityInstance;
 
     @Override
     public <V> ServerCallContext create(StreamObserver<V> responseObserver) {
--- a/compat-0.3/reference/grpc/src/main/java/org/a2aproject/sdk/compat03/server/grpc/quarkus/QuarkusCallContextFactory_v0_3.java
+++ b/compat-0.3/reference/grpc/src/main/java/org/a2aproject/sdk/compat03/server/grpc/quarkus/QuarkusCallContextFactory_v0_3.java
@@ -1,0 +1,78 @@
+package org.a2aproject.sdk.compat03.server.grpc.quarkus;
+
+import static org.a2aproject.sdk.server.ServerCallContext.TRANSPORT_KEY;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import jakarta.enterprise.context.ApplicationScoped;
+import jakarta.enterprise.inject.Instance;
+import jakarta.inject.Inject;
+
+import io.grpc.Context;
+import io.grpc.Metadata;
+import io.grpc.stub.StreamObserver;
+import io.quarkus.security.identity.SecurityIdentity;
+import org.a2aproject.sdk.compat03.conversion.A2AProtocol_v0_3;
+import org.a2aproject.sdk.compat03.transport.grpc.context.GrpcContextKeys_v0_3;
+import org.a2aproject.sdk.compat03.transport.grpc.handler.CallContextFactory_v0_3;
+import org.a2aproject.sdk.server.ServerCallContext;
+import org.a2aproject.sdk.server.auth.AuthenticatedUser;
+import org.a2aproject.sdk.server.auth.UnauthenticatedUser;
+import org.a2aproject.sdk.server.auth.User;
+import org.a2aproject.sdk.spec.TransportProtocol;
+
+@ApplicationScoped
+public class QuarkusCallContextFactory_v0_3 implements CallContextFactory_v0_3 {
+
+    @Inject
+    Instance<SecurityIdentity> securityIdentityInstance;
+
+    @Override
+    public <V> ServerCallContext create(StreamObserver<V> responseObserver) {
+        User user;
+        if (securityIdentityInstance.isResolvable()) {
+            SecurityIdentity securityIdentity = securityIdentityInstance.get();
+            if (!securityIdentity.isAnonymous()) {
+                user = new AuthenticatedUser(securityIdentity.getPrincipal().getName());
+            } else {
+                user = UnauthenticatedUser.INSTANCE;
+            }
+        } else {
+            user = UnauthenticatedUser.INSTANCE;
+        }
+
+        Map<String, Object> state = new HashMap<>();
+        state.put(TRANSPORT_KEY, TransportProtocol.GRPC);
+        state.put("grpc_response_observer", responseObserver);
+
+        Context currentContext = Context.current();
+        if (currentContext != null) {
+            state.put("grpc_context", currentContext);
+            io.grpc.Metadata grpcMetadata = GrpcContextKeys_v0_3.METADATA_KEY.get(currentContext);
+            if (grpcMetadata != null) {
+                state.put("grpc_metadata", grpcMetadata);
+                Map<String, String> headers = new HashMap<>();
+                for (String key : grpcMetadata.keys()) {
+                    if (key.endsWith(Metadata.BINARY_HEADER_SUFFIX)) {
+                        continue;
+                    }
+                    headers.put(key, grpcMetadata.get(Metadata.Key.of(key, Metadata.ASCII_STRING_MARSHALLER)));
+                }
+                state.put("headers", headers);
+            }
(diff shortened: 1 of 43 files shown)


server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryTaskStore.java
```diff
@@ -123,6 +146,16 @@ public ListTasksResult list(ListTasksParams params) {
...
 
+        if (authorizationProvider != null && context != null) {
+            List<Task> authorizedTasks = new ArrayList<>(allFilteredTasks.size());
+            for (Task task : allFilteredTasks) {
+                if (authorizationProvider.checkRead(context, task.id(), TaskOperation.LIST_TASKS)) {
+                    authorizedTasks.add(task);
+                }
+            }
+            allFilteredTasks = authorizedTasks;
+        }
```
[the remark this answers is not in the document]
kabir: @gemini-code-assist A2AError extends RuntimeException so it is unchecked

extras/task-store-database-jpa/src/main/java/org/a2aproject/sdk/extras/taskstore/database/jpa/JpaDatabaseTaskStore.java
```diff
@@ -250,109 +268,74 @@ public boolean isTaskFinalized(String taskId) {
...
+
+                    dbExhausted = batch.size() < limit;
+                    int batchEnd = Math.min(batch.size(), remaining);
+
+                    for (int i = 0; i < batchEnd; i++) {
+                        Task task = deserializeTask(batch.get(i));
+                        if (authorizationProvider.checkRead(context, task.id(), TaskOperation.LIST_TASKS)) {
+                            tasks.add(task);
+                        }
+                    }
```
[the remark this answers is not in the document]
kabir: @gemini-code-assist A2AError extends RuntimeException so it is unchecked

compat-0.3/reference/grpc/src/main/java/org/a2aproject/sdk/compat03/server/grpc/quarkus/QuarkusCallContextFactory_v0_3.java
```diff
@@ -0,0 +1,87 @@
...
+@ApplicationScoped
+public class QuarkusCallContextFactory_v0_3 implements CallContextFactory_v0_3 {
+
+    @Inject
+    Instance<SecurityIdentity> securityIdentityInstance;
+
+    @Override
+    public <V> ServerCallContext create(StreamObserver<V> responseObserver) {
+        User user;
+        if (securityIdentityInstance.isResolvable()) {
```
ehsavoie, line 35: This looks like a duplicate of QuarkusCallContextFactory maybe we should have a helper method for this

server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryTaskStore.java
```diff
@@ -123,6 +146,16 @@ public ListTasksResult list(ListTasksParams params) {
                         .thenComparing(Task::id))
                 .toList();
 
+        if (authorizationProvider != null && context != null) {
```
ehsavoie: Why not adding this in the stream filtering above ?

extras/task-store-database-jpa/src/main/java/org/a2aproject/sdk/extras/taskstore/database/jpa/JpaDatabaseTaskStore.java
```diff
@@ -50,6 +56,19 @@ public class JpaDatabaseTaskStore implements TaskStore, TaskStateProvider {
     @Inject
     A2AConfigProvider configProvider;
 
+    private final @org.jspecify.annotations.Nullable TaskAuthorizationProvider authorizationProvider;
```
ehsavoie: Should be @Nullable

ehsavoie (changes requested):
The User interface only has isAuthenticated() and getUsername(). There's no way to get roles or claims. The TaskAuthorizationProvider receives a ServerCallContext so implementers can reach into state to get the raw SecurityIdentity, but this couples the provider to Quarkus  internals. A getRoles(): Set<String> method or getAttribute(String): Object would make the interface more useful for role-based authorization without requiring framework coupling. This isn't blocking, but it's a natural next step.

The anonymous User implementations in QuarkusCallContextFactory. Lines 39-49 create an anonymous User implementation inline. This works but makes it impossible to test equality or do instanceof checks. A named record like AuthenticatedUser(String username) would be cleaner and more testable.

server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/AuthorizationRequestHandlerDecorator.java
```diff
@@ -0,0 +1,253 @@
...
+                    authorizationProvider);
+        }
+        return publisher;
+    }
+
+    @Override
+    public TaskPushNotificationConfig onCreateTaskPushNotificationConfig(TaskPushNotificationConfig params,
+            ServerCallContext context) throws A2AError {
+        String taskId = params.taskId();
+        if (taskId != null) {
```
ehsavoie, line 215: If taskId is null the task can be created ? Is that correct ?

server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/AuthorizationRequestHandlerDecorator.java
```diff
@@ -0,0 +1,253 @@
...
+    public EventKind onMessageSend(MessageSendParams params, ServerCallContext context) throws A2AError {
+        String taskId = params.message().taskId();
+        if (taskId != null) {
+            enforceWrite(context, taskId, TaskOperation.MESSAGE_SEND);
+        } else {
+            enforceCreate(context, TaskOperation.MESSAGE_SEND);
+        }
+        EventKind result = delegate.onMessageSend(params, context);
+        String resultTaskId = extractTaskId(result);
+        recordOwnershipIfNeeded(context, resultTaskId, TaskOperation.MESSAGE_SEND);
```
ehsavoie, line 190: TOCTOU race in onMessageSend ownership recording.

If two concurrent onMessageSend calls arrive for the same new task (no taskId in the message), both pass enforceCreate, the delegate creates the task, and both reach recordOwnershipIfNeeded. The isTaskRecorded + recordOwnership sequence isn't atomic. The javadoc says recordOwnership "must be idempotent," which handles the data integrity side, but there's a window where isTaskRecorded returns false for both threads. The TestTaskAuthorizationProvider correctly uses putIfAbsent, so this is safe in practice as long as implementers follow the
  idempotency contract. This is documented but worth highlighting in a "common pitfall" section of the interface javadoc.

server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/AuthorizationRequestHandlerDecorator.java
kabir: No — if taskId is null, the delegate (DefaultRequestHandler) immediately throws InvalidParamsError("taskId is required") at line 793. The null guard in the  decorator just avoids calling enforceWrite with a null argument; the request is never allowed through.

extras/task-store-database-jpa/src/main/java/org/a2aproject/sdk/extras/taskstore/database/jpa/JpaDatabaseTaskStore.java
```diff
@@ -250,109 +269,74 @@ public boolean isTaskFinalized(String taskId) {
...
+                            tasks.add(task);
+                        }
+                    }
+
+                    // Advance cursor to last fetched DB row for next iteration
+                    if (batchEnd > 0) {
+                        JpaTask last = batch.get(batchEnd - 1);
+                        cursor = new PageToken(last.getStatusTimestamp(), last.getId());
+                    }
+                }
```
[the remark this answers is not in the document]
kabir: @gemini-code-assist A2AError is NOT checked

server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryTaskStore.java
```diff
@@ -114,6 +136,8 @@ public ListTasksResult list(ListTasksParams params) {
                         (task.status() != null &&
                          task.status().timestamp() != null &&
                          task.status().timestamp().toInstant().isAfter(params.statusTimestampAfter())))
+                .filter(task -> authorizationProvider == null || context == null ||
+                        authorizationProvider.checkRead(context, task.id(), TaskOperation.LIST_TASKS))
```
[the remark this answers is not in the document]
kabir: @gemini-code-assist A2AError is not checked

PR #937, opened 2026-06-17
fix: add no-args constructors to response/request classes for Gson compa…

kabir:
…tibility

Gson's reflective adapter falls back to sun.misc.Unsafe.allocateInstance() when no no-args constructor is available. This fails in modular environments like WildFly's JBoss Modules. Adding no-args constructors lets Gson instantiate these classes directly without Unsafe.


ehsavoie (approved):

ehsavoie:
Thanks @kabir

PR #966, opened 2026-07-01
feat: expose programmatic auth wiring

### LINKED ISSUE #965: [Feat]: Expose programmatic authorization wiring for non-CDI runtimes

The A2A Java SDK already provides task-level authorization through `TaskAuthorizationProvider` and `AuthorizationRequestHandlerDecorator`.

This works well in CDI runtimes, where the container wires decorators automatically.

However, non-CDI integrations (for example Spring Boot) create `DefaultRequestHandler` and `TaskStore` programmatically, which makes it impossible to reuse the existing authorization flow because the framework-neutral constructors are not public.

Expose the programmatic APIs by making these constructors public:

- `AuthorizationRequestHandlerDecorator(RequestHandler, TaskAuthorizationProvider)`
- `InMemoryTaskStore(TaskAuthorizationProvider)`

Additionally, it would be useful to allow `AuthenticatedUser` to carry optional attributes (e.g. roles, scopes, tenant, clientId) so integrations can pass richer identity context into `TaskAuthorizationProvider`.

## Why

This would allow framework integrations to reuse the existing SDK authorization model without introducing framework-specific code or duplicating authorization logic.

_No response_

### Additional context

_No response_

### Code of Conduct

Sh1bari:
Closes #965

## Summary

This PR exposes the existing authorization wiring APIs for non-CDI runtimes.

The SDK already provides task-level authorization through `TaskAuthorizationProvider` and `AuthorizationRequestHandlerDecorator`, but non-CDI integrations (such as Spring Boot) create `DefaultRequestHandler` and `TaskStore` programmatically and cannot reuse the existing authorization flow because the framework-neutral constructors are not public.

## Changes

- make `AuthorizationRequestHandlerDecorator(RequestHandler, TaskAuthorizationProvider)` public
- make `InMemoryTaskStore(TaskAuthorizationProvider)` public
- add optional attributes support to `AuthenticatedUser`
- add common user attribute constants

## Why

This allows framework integrations to reuse the existing SDK authorization model without introducing framework-specific code or duplicating authorization logic.


server-common/src/main/java/org/a2aproject/sdk/server/auth/UserAttributes.java
```diff
@@ -0,0 +1,12 @@
+package org.a2aproject.sdk.server.auth;
+
+public final class UserAttributes {
+
+    public static final String AUTHORITIES = "authorities";
```
kabir: What are these for? They seem to be unused. 

I guess they are meant to be used as lookups in AuthenticatedUser.attributes, but do we really know the full/expected set of attributes?

My vote is to not have these constants here, but instead push that to the integrators wanting to add extra stuff

kabir (changes requested):

server-common/src/main/java/org/a2aproject/sdk/server/auth/UserAttributes.java
Sh1bari: Removed the file. I thought these would be base attributes for extending User and handling auth, but you're right — better to keep this controlled externally.

kabir (approved):

kabir:
Thanks!

PR #969, opened 2026-07-04
fix: PR checks fail after actions/checkout@v6 -> v7

Sh1bari:
Summary:
- Replaced the previous workflow_run-based release profile check with a regular workflow that runs directly on pull_request, push to main, and workflow_dispatch.
- PRs now verify the build with `-P release`, but without access to GPG keys or Maven Central credentials.
- Added safe PR-check parameters: `-Dgpg.skip=true` and `-Drelease.auto.publish=false`.

Why:
After upgrading `actions/checkout` to v7, checking out fork PR code from a privileged workflow_run context is blocked as unsafe. The old workflow executed untrusted PR code while repository secrets were available, which created a pwn-request risk.

This change keeps release profile validation for PRs while making it safe for fork PRs. Real GPG signing and Maven Central publishing are still verified by the release workflow during an actual release.


kabir:
Actually, the point of this job is to make sure that the javadoc generation and secrets etc. work so that we don't get any nasty surprises when we do the release.

I do see the job failing though when I tried with a test PR.

I'd like to keep this if possible, so I will look into this.

kabir:
Superseded by https://github.com/a2aproject/a2a-java/pull/972

PR #974, opened 2026-07-07
fix!: TaskState.UNRECOGNIZED is called TASK_STATE_UNSPECIFIED in spec

### LINKED ISSUE #958: [Bug]: Update TaskState enum UNRECOGNIZED to TASK_STATE_UNSPECIFIED
### What happened?

Hi, I am looking at v1.1.0.Final version.

I found that the latest version of the org.a2aproject.sdk.spec.TaskState enum class still has value UNRECOGNIZED  instead of TASK_STATE_UNSPECIFIED, which is inconsistent with the [a2a-protocol](https://a2a-protocol.org/latest/specification/#413-taskstate).

Please fix this so that we don't have to do the conversion and validation ourselves. 

### Relevant log output

```shell

```

### Code of Conduct

kabir:
BREAKING CHANGE: This is an actual bug in our implementation so it was seen as suitable for a micro release. Users compiling against UNRECOGNIZED will need to replace this with TASK_STATE_UNSPECIFIED.

Fixes #958 🦕

### DIFF
--- a/client/base/src/main/java/org/a2aproject/sdk/client/ClientTaskManager.java
+++ b/client/base/src/main/java/org/a2aproject/sdk/client/ClientTaskManager.java
@@ -60,7 +60,7 @@ class ClientTaskManager {
         Task task = currentTask;
         if (task == null) {
             task = Task.builder()
-                    .status(new TaskStatus(TaskState.UNRECOGNIZED))
+                    .status(new TaskStatus(TaskState.TASK_STATE_UNSPECIFIED))
                     .id(taskId)
                     .contextId(contextId == null ? "" : contextId)
                     .build();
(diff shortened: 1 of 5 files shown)


wushunhao (changes requested):
It looks good, but I found another inconsistency. 

According to [4.1.3. TaskState](https://a2a-protocol.org/latest/specification/#413-taskstate), TASK_STATE_UNSPECIFIED is not described as a terminal state. 

But I guess it also makes sense to terminate the task when it's in this state. What do you think?

kabir:
@wushunhao I'll try to make it non-terminal and see what happens on CI

jmesnil (approved):

PR #976, opened 2026-07-09
fix: ensure correct task/contextId in emitted Messages

kabir:


server-common/src/main/java/org/a2aproject/sdk/server/tasks/AgentEmitter.java
```diff
@@ -508,6 +508,14 @@ public void sendMessage(List<Part<?>> parts, @Nullable Map<String, Object> metad
      * @since 1.0.0
      */
     public void sendMessage(Message message) {
+        if (message.taskId() != null && !message.taskId().equals(taskId)) {
+            throw new IllegalArgumentException(
+                    String.format("Message taskId mismatch: expected=%s, actual=%s", taskId, message.taskId()));
```
jmesnil: I'm not sure about this... We might want to avoid "leaking" out the task `taskId` to avoid further forgery

server-common/src/main/java/org/a2aproject/sdk/server/tasks/AgentEmitter.java
```diff
@@ -508,6 +508,14 @@ public void sendMessage(List<Part<?>> parts, @Nullable Map<String, Object> metad
...
+        }
+        if (message.contextId() != null && !message.contextId().equals(contextId)) {
+            throw new IllegalArgumentException(
+                    String.format("Message contextId mismatch: expected=%s, actual=%s", contextId, message.contextId()));
```
jmesnil: idem

jmesnil (changes requested):

server-common/src/main/java/org/a2aproject/sdk/server/tasks/AgentEmitter.java
kabir: @jmesnil Fixed. Removed from error message, and adding error logging instead

server-common/src/main/java/org/a2aproject/sdk/server/tasks/AgentEmitter.java
kabir: Fixed as above

jmesnil (approved):

PR #756, opened 2026-03-17
fix: Keep stream open on interrupted state changes

kabir:
Fixes #754

PR #980, opened 2026-07-13
fix(server): close event stream when requiresInput/requiresAuth force isFinal

### LINKED ISSUE #975: [Bug]: AgentEmitter#requiresInput(..., isFinal=true) does not close the event stream
### Describe the bug

After #756, streams correctly stay open for interrupted states (`INPUT_REQUIRED` / `AUTH_REQUIRED`).

However, `AgentEmitter#requiresInput(Message, boolean isFinal)` and `requiresAuth(..., boolean isFinal)` accept an `isFinal` flag that appears to have no effect on stream completion. Passing `isFinal=true` still leaves the SSE / `Flow.Publisher` open (subscriber `onComplete` is never called).

### Expected behavior

Either:
1. `requiresInput(message, true)` / `requiresAuth(message, true)` closes the stream after the status event is delivered, **or**
2. The `isFinal` parameter is documented as not controlling stream lifecycle (and ideally removed/deprecated if unused).

### Actual behavior

In `1.1.0.Final`, `isFinal` only sets the internal `terminalStateReached` flag in `AgentEmitter#updateStatus`. It is **not** written onto `TaskStatusUpdateEvent`.

`EventConsumer` terminates the stream when `TaskStatusUpdateEvent#isFinal()` is true, but that method only delegates to `TaskState#isFinal()`, which is always `false` for `INPUT_REQUIRED` / `AUTH_REQUIRED`.

So `requiresInput(msg, true)` and `requiresInput(msg, false)` behave the same for stream completion.

Related leftover: `AgentEmitter#emitEvent` Javadoc still shows `TaskStatusUpdateEvent.builder()...isFinal(false)`, but the builder no longer has `isFinal(...)`.

### Steps to reproduce

1. In an `AgentExecutor`, after work finishes for this turn, call:
   ```java
   emitter.requiresInput(replyMessage, true);
   ```
2. Subscribe to `onMessageSendStream` / JSON-RPC streaming response.
3. Observe that events are delivered, but the publisher never completes / SSE connection stays open.

### Version

- a2a-java SDK: `1.1.0.Final`
- Related change: #756 / #754

### Question

If keeping the stream open for interrupted states is intentional (per #756), what is the supported way for an agent to end the current stream after emitting `INPUT_REQUIRED`, when the client is expected to continue via a new request rather than the same open stream?

### Relevant log output

```shell

```

### Code of Conduct

chopmob-cloud:
Fixes #975.

### Problem
`AgentEmitter#requiresInput(message, isFinal=true)` (and `requiresAuth(..., isFinal=true)`) set the internal `terminalStateReached` flag but never completed the stream. The emitted `TaskStatusUpdateEvent#isFinal()` is derived from the task state (`status.state().isFinal()`), which is `false` for the interrupted `INPUT_REQUIRED` / `AUTH_REQUIRED` states, so `EventConsumer` kept the SSE / `Flow.Publisher` open. The `isFinal` argument therefore only blocked further updates; it could not end the stream, and connections an agent asked to release stayed open.

### Fix
When the caller forces finality on a non-terminal state, enqueue the existing `QueueClosedEvent` after the status event:

```java
if (isFinal && !state.isFinal()) {
    eventQueue.enqueueEvent(new QueueClosedEvent(taskId));
}
```

`EventConsumer` already treats `QueueClosedEvent` as a stream-terminating event, so the status event is delivered and then the stream completes, letting the client resume via a new request. This uses the SDK's own close path and touches neither the event record nor the proto. Interrupted states still stay open by default (#756); only an explicit `isFinal=true` closes them. Terminal states are unaffected (their event's `isFinal()` is already `true`, and the `!state.isFinal()` guard skips them).

Also removes a stale `.isFinal(false)` call from the `emitEvent` Javadoc, since the builder no longer exposes it.

### Tests
- Updated the four interrupted+final tests to assert the trailing `QueueClosedEvent` via a backward-compatible `expectStreamClose` overload of the existing helper.
- Verified on JDK 17: `server-common` suite 412/412 pass; with the fix reverted, the four forced-final tests fail (confirming they exercise the behavior).

### Note on approach
This implements the "wire `isFinal` to actually close the stream" option. If you'd rather treat interrupted streams as always staying open and have `isFinal` on `requiresInput`/`requiresAuth` deprecated/documented instead, I'm happy to switch to that.


kabir:
Hi, sorry for the delay, there is a lot going on this week, and I'm on my own.

I looked at the code, since I didn't remember what 'forcing final' meant, and then I saw the `AgentEmitter.requiresInput()` and `requiresAuth()` methods have an overload with the extra `isFinal` parameter.

I don't remember adding that, or why I added it :-)

I asked Claude to check if this is something coming from the spec, and it says:

> exactly these fields for TaskStatus:
> 
>   - state (TaskState, required)
>   - message (Message)
>   - timestamp (Timestamp)
> 
>   And TaskStatusUpdateEvent has:
> 
>   - task_id, context_id, status, metadata
> 
>   No boolean final anywhere.
> 
>   The spec also clearly distinguishes:
>   - Terminal states: COMPLETED, FAILED, CANCELED, REJECTED
>   - Interrupted states: INPUT_REQUIRED, AUTH_REQUIRED
> 
>   Interrupted states are explicitly not terminal — they pause execution awaiting further input/auth.
> 
>   So the isFinal parameter on requiresInput() and requiresAuth() in AgentEmitter is a SDK invention with no basis in the
>   spec. It lets you mark an interrupted state as "final" (preventing further status updates), which contradicts the
>   spec's model where interrupted states are resumable.
> 
>   Looks like it should be removed. Want me to clean that up?

So I wonder now if this was something which was accidentally added at some point.
Let's see what @ehsavoie and @jmesnil say when they come back.

chopmob-cloud:
Thanks for taking a look, and no worries on timing.

I think the confusion is worth naming directly: `isFinal` here controls the stream/connection lifecycle, not the task state. Those are independent in A2A. `INPUT_REQUIRED` / `AUTH_REQUIRED` stay non-terminal and resumable in both cases, and `TaskState#isFinal()` is correctly `false` throughout, which this PR does not change. What `isFinal=true` decides is whether this SSE / `Flow.Publisher` stays open (client resumes on the same stream, the #756 default) or completes (client resumes via a new `message/send` or `tasks/resubscribe`). Closing the stream on an interrupted state is just the agent releasing the connection while it waits, not asserting a terminal task state. Before the fix it set the internal flag but never actually closed the stream, so the option was a no-op; this makes the close path real via the SDK's existing `QueueClosedEvent`.

That said, #975 from @bianjp listed both "make it close" and "remove/deprecate if unused" as acceptable, so the API-surface call is genuinely yours. If you would rather drop the `isFinal` overloads to keep the surface minimal, that also resolves #975 and I am happy to close this PR. If the "release the connection on an interrupted state" capability is worth keeping, this makes it behave as documented. Either way it resolves the reported bug, and I am glad to follow whichever direction you and @ehsavoie / @jmesnil settle on.

kabir:
Hi, 

Thank you for the analysis and the PR.

I've finally had some time to look more into this, and found the following.

The `isFinal` parameter on `requiresInput()` and `requiresAuth()` isn't something that comes from the spec. 

The spec is pretty clear that finality is derived from the task state itself — terminal states (`COMPLETED`, `FAILED`, `CANCELED`, `REJECTED`) close the stream, and interrupted states (`INPUT_REQUIRED`, `AUTH_REQUIRED`) don't. Section 7.6.1 even says agents "SHOULD maintain any active response streams" after `AUTH_REQUIRED`.

I also checked the Python SDK — `TaskUpdater.requires_input()` and `requires_auth()` there don't have an isFinal parameter at all.

(Side note: the Python SDK's EventConsumer does treat `INPUT_REQUIRED` as stream-closing but not `AUTH_REQUIRED`, which looks like its own inconsistency  there — but that's a separate problem.)

So the `isFinal` overloads look like a Java SDK invention that predates #756, and I don't think we should be adding more logic to support something that isn't in the spec. I also can't figure out why I added them!

Instead I think I would like to:
- Deprecate the four `isFinal` overloads for `requiresInput` and `requiresAuth`)
- Have them just delegate to the non-`isFinal` versions (effectively ignoring the parameter — which is close to what
  happens today anyway), and document in the Javadoc that the parameters are ignored
- Remove them in a future major version

If we ever need a way for agents to explicitly close a stream on an interrupted state, that should probably be a dedicated, well-named method rather than a boolean flag that blurs the line between task state and connection lifecycle.

Does this sound reasonable? From your comments, I think it sounds like you're happy to go either way?

I have opened #989 as an alternative PR

chopmob-cloud:
No worries, good luck with it!

PR #991, opened 2026-07-23
fix(server): reconcile blocking result with TaskStore

### LINKED ISSUE #2049: [Bug]: Blocking onMessageSend can throw "Could not find a Task/Message" even when TaskStore already has the completed Task
### What happened?

Under concurrent blocking SendMessage / onMessageSend load, the Java server SDK can throw:

InternalError: Could not find a Task/Message for {taskId}
  at ResultAggregator.consumeAndBreakOnInterrupt

even when the same taskId is already persisted in TaskStore (COMPLETED).

This is not a TaskStore primary-key miss. Persistence often succeeds; sync response assembly fails earlier.

SDK version

- a2a-java-sdk-server-common 1.0.0.CR1 (also relevant for current mainline ResultAggregator / DefaultRequestHandler shape)

Observed behavior

In DefaultRequestHandler.onMessageSend, the blocking follow-up path is documented in the SDK source as these steps:

Step 1: Wait for the agent future to finish enqueueing events (or timeout).
Step 2: Close the queue so EventConsumer can finish.
Step 3: Wait for background consumption to complete.
Step 4: Rely on MainEventBusProcessor ordering (persist before distribute) so consumed events are already in TaskStore.
Step 5: Call taskStore.get(taskId) and use that Task as the returned result.

Important: those Steps 1-5 only run after ResultAggregator.consumeAndBreakOnInterrupt returns successfully with some EventKind.

Failure path

- Agent emits events.
- MainEventBusProcessor may already have persisted the Task to TaskStore.
- ResultAggregator.consumeAndBreakOnInterrupt then tries, in order:
  1) captured Message
  2) captured Task event
  3) taskManager.getTask()
  and if all are null, throws "Could not find a Task/Message for {taskId}".
- Because that throw happens inside the aggregator, DefaultRequestHandler never enters the blocking follow-up path, so Step 5 (taskStore.get(taskId)) never runs.

Clients therefore get InternalError even when TaskStore already has the Task.

Why this looks like an SDK gap

1. Interrupt before a returnable kind exists
   For blocking calls, aggregator sets shouldInterrupt=true on events (including non-Task / non-Message events such as status updates). It only stores return values for Message and Task. The waiting thread can unblock with message == null, capturedTask == null, and taskManager.getTask() == null, then throw, while TaskStore may already have the Task.

2. Misleading fallback
   Comment near the end of consumeAndBreakOnInterrupt implies a TaskStore fetch, but the code uses taskManager.getTask(), not taskStore.get(taskId). After TaskStore updates moved to MainEventBusProcessor, TaskManager may not be hydrated, so the fallback is weak.

3. TaskStore refresh is only Step 5
   The durable re-read is intentionally in DefaultRequestHandler Step 5. Empty aggregator capture throws first, so Step 5 is skipped.

Expected behavior

If in-memory capture is empty for a known taskId, blocking onMessageSend should:
- call taskStore.get(taskId) (and/or wait briefly for persistence), and
- return that Task when present,
- only fail if TaskStore also has no row.

Alternatively: do not complete/interrupt the blocking wait until a Task or Message (or hydrated TaskManager task) is available.

Reproduction sketch

- Host using DefaultRequestHandler with a TaskStore and in-memory event queues
- Concurrent blocking SendMessage under parallel load
- Intermittent InternalError with the message above
- A later TaskStore.get(taskId) returns a COMPLETED Task for the same id

Happy to add a minimal unit/integration repro if helpful.

Workaround

Host-side: catch this specific InternalError, parse taskId, poll TaskStore.get(taskId) for a short bound, return Task if found.

Related (not duplicates)

- #428 blocking until final state
- #430 / #441 wait for agent + consumption in blocking path
- #780 EventConsumer race fixes

Those address related timing issues; we did not find an open issue for "aggregator empty -> throw despite TaskStore hit."

Environment

- Language: Java
- Artifact: a2a-java-sdk-server-common
- Mode: blocking non-streaming onMessageSend

### Relevant log output

```shell

```

### Code of Conduct

jstar0:
## Summary

Fixes a race in the blocking Java server response path tracked by [a2aproject/A2A#2049](https://github.com/a2aproject/A2A/issues/2049). `onMessageSend` can currently return an internal error when the agent is complete but TaskStore persistence is still becoming visible to the response aggregator.

## Root Cause / Context

When the event consumer completes without delivering a Message or full Task event, `ResultAggregator` performs one TaskStore lookup and immediately raises `Could not find a Task/Message` if the row is not visible yet. That prevents the later response-handler TaskStore reconciliation from running.

## Changes

- Poll TaskStore with a bounded timeout when a blocking call has no in-memory result, while preserving the single-read behavior for non-blocking calls.
- Make the reconciliation timeout configurable through `a2a.blocking.reconciliation.timeout.seconds`, retaining the existing one-second behavior by default.
- Add debug logging for reconciliation attempts and successful delayed lookups.
- Document the new setting and add regression coverage for delayed visibility, permanent absence, non-blocking behavior, configuration wiring, and the default property value.

## Scope / Risk

Only the blocking empty-capture path polls TaskStore, and the default timeout remains one second. EventConsumer lifecycle, TaskStore APIs, cancellation, streaming emission, and the public A2A protocol are unchanged.

## Verification

```bash
mvn -pl server-common test
mvn -pl integrations/microprofile-config test
mvn -pl server-common,integrations/microprofile-config -am -Dtest=ResultAggregatorTest,DefaultRequestHandlerTest,MicroProfileConfigProviderTest -Dsurefire.failIfNoSpecifiedTests=false test
git diff --check
```

The module suites pass 416 `server-common` tests and 10 MicroProfile config tests. The focused cross-module run passes 46 tests.

Fixes a2aproject/A2A#2049

### DIFF
--- a/server-common/src/main/java/org/a2aproject/sdk/server/tasks/ResultAggregator.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/tasks/ResultAggregator.java
@@ -8,6 +8,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Flow;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -28,6 +29,8 @@ import org.slf4j.LoggerFactory;
 
 public class ResultAggregator {
     private static final Logger LOGGER = LoggerFactory.getLogger(ResultAggregator.class);
+    private static final long TASK_STORE_RECONCILIATION_TIMEOUT_NANOS = TimeUnit.SECONDS.toNanos(1);
+    private static final long TASK_STORE_RECONCILIATION_POLL_MILLIS = 10;
 
     private final TaskManager taskManager;
     private final Executor executor;
@@ -235,7 +238,7 @@ public class ResultAggregator {
             Utils.rethrow(error);
         }
 
-        // Return Message if captured, otherwise Task if captured, otherwise fetch from TaskStore
+        // Return Message if captured, otherwise Task if captured, otherwise reconcile with TaskStore.
         EventKind eventKind = message.get();
         if (eventKind == null) {
             eventKind = capturedTask.get();
@@ -244,7 +247,7 @@ public class ResultAggregator {
             }
         }
         if (eventKind == null) {
-            eventKind = taskManager.getTask();
+            eventKind = reconcileTaskStore(blocking);
             if (LOGGER.isDebugEnabled() && eventKind instanceof Task t) {
                 LOGGER.debug("Returning task from TaskStore: id={}, state={}", t.id(), t.status().state());
             }
@@ -257,6 +260,29 @@ public class ResultAggregator {
                 eventKind,
                 interrupted.get(),
                 consumptionCompletionFuture);
+    }
+
+    private @Nullable Task reconcileTaskStore(boolean blocking) throws A2AError {
+        Task task = taskManager.getTask();
+        if (task != null || !blocking) {
+            return task;
+        }
+
+        long deadline = System.nanoTime() + TASK_STORE_RECONCILIATION_TIMEOUT_NANOS;
+        while (System.nanoTime() < deadline) {
+            try {
+                Thread.sleep(TASK_STORE_RECONCILIATION_POLL_MILLIS);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new InternalError("Interrupted while reconciling TaskStore for " + taskManager.getTaskId());
+            }
+
+            task = taskManager.getTask();
+            if (task != null) {
+                return task;
+            }
+        }
+        return null;
     }
 
     private String taskIdForLogging() {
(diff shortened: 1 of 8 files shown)


kabir:
@jstar0 Thank you! I've opened https://github.com/jstar0/a2a-java/pull/1 with a few fixes to make the timeout configurable and some debug logging.

jstar0:
Thanks @kabir. I integrated your configurable timeout and debug logging commit, then added coverage for the configuration wiring and default value. The focused 46-test run and the complete `server-common` (416) and MicroProfile config (10) suites pass.

server-common/src/main/java/org/a2aproject/sdk/server/tasks/ResultAggregator.java
```diff
@@ -37,6 +38,12 @@ public class ResultAggregator {
     private final long reconciliationTimeoutNanos;
     private volatile @Nullable Message message;
 
+    public ResultAggregator(TaskManager taskManager, @Nullable Message message, Executor executor,
```
kabir: I"m not sure I see the point of readding this constructor.

Although the class is public the usage is purely internal, in `DefaultRequestHandler` and the tests. So I don't think It is needed.

kabir (changes requested):
I don't se the need for the re-introduced constructor. Assuming I'm not mistaken, please get rid of it.
The additional tests look good though 👍

server-common/src/main/java/org/a2aproject/sdk/server/tasks/ResultAggregator.java
jstar0: Agreed. I removed the four-argument overload and updated the tests to use the configurable constructor. The focused 36-test run and the complete `server-common` suite (416 tests) pass.

ehsavoie (approved):

ehsavoie:
thanks @jstar0

PR #992, opened 2026-07-23
test: build itk instruction proto straight from repository and remove local copy

JakubWorek:
# Description

Adjusts itk pipeline to use instruction proto straight from cloned itk repository during startup of itk test


.github/workflows/build-and-test.yml
```diff
@@ -34,7 +34,11 @@ jobs:
         distribution: 'temurin'
         cache: maven
     - name: Build with Maven and run tests
-      run: mvn -B package --file pom.xml -fae ${{ matrix.os == 'windows-latest' && '-DskipDockerTests=true' || '' }}
+      # The `itk` module is excluded here because its proto sources live in
+      # the external `a2a-itk` repo and are fetched at test time by
+      # `itk/run_itk.sh`. It is exercised by the dedicated `itk.yaml` and
+      # `itk-nightly.yaml` workflows instead.
+      run: mvn -B package --file pom.xml -fae -pl '!itk' ${{ matrix.os == 'windows-latest' && '-DskipDockerTests=true' || '' }}
```
kabir: @JakubWorek I wonder if instead of having to specify `-pl !itk` in all the GH Actions workflow files, if it might be better to change the parent pom so that the itk module is included in the list of modules, and have a 'itk' profile to enable it?

That will save adjusting all the jobs. We also have other repositories consuming this module, which in some cases need to build a SNAPSHOT of a2a-java, so this will have knock-on effects

kabir (changes requested):

.github/workflows/build-and-test.yml
ehsavoie: I agree with @kabir: Why not create a itk profile disabled by default in the parent pom that would add the itk module ? So you just enable it in the ITK workflow

JakubWorek: ack, I'll adjust the solution

itk/pom.xml
```diff
@@ -98,8 +98,45 @@
...
 
+    <properties>
+        <!--
+          Single-source location of the canonical .proto files. The a2a-itk
+          repository is fetched at the ref pinned by A2A_ITK_REVISION (see
+          run_itk.sh) into ${project.basedir}/a2a-itk. Override via
+          -Da2a.itk.proto.dir=... when building outside run_itk.sh.
+        -->
+        <a2a.itk.proto.dir>${project.basedir}/a2a-itk/protos</a2a.itk.proto.dir>
+        <protobuf-maven-plugin.version>4.1.2</protobuf-maven-plugin.version>
```
ehsavoie: This is now managed by the parent pom

ehsavoie (changes requested):

.github/workflows/build-and-test.yml
JakubWorek: done, now itk pipeline wont be green until https://github.com/a2aproject/a2a-itk/pull/26 lands

itk/pom.xml
JakubWorek: thank you, I'll adjust to latest main branch

ehsavoie (approved):

PR #993, opened 2026-07-23
chore: upgrade to Quarkus 3.37.3 and align dependency versions

ehsavoie:
- Quarkus 3.36.3 -> 3.37.3
- protobuf-java 4.33.2 -> 4.35.0 (required by gRPC 1.81.0 gencode)
- JUnit 6.0.3 -> 6.1.1
- rest-assured 6.0.0 -> 6.0.1
- JBang example: OpenTelemetry SDK 1.55.0 -> 1.60.1, gRPC Netty 1.77.0 -> 1.81.0


kabir:
Does the protobuf upgrade work with the gRPC version in WildFly, for A2A-jakarta?

ehsavoie:
@kabir yes, I wanted your feedback there :)

hpehl:
@ehsavoie the gRPC feature pack is ready to be released. 

These are the current dependency versions: 
https://github.com/wildfly-extras/wildfly-grpc-feature-pack/blob/main/pom.xml#L56

Does that work for you?

kabir (commented):
I've verified that a2a-jakarta works with these changes, and the grpc feature pack changes.
But I think we should hold off on merging this until we have the grpc feature pack release

kabir (approved):
Actually, it seems a2a-jakarta works fine against this PR without the WildFly grpc feature pack upgrade.

My guess is that as the protobuf/grpc classes were generated with an older version, it runs fine on the version that the WildFly grpc feature pack has.
And I think once released, we don't regenerate the generated classes with newer versions.
When the next version of the spec is out, we will use whatever grpc/protobuf version we have installed to generate the classes for that version

PR #1005, opened 2026-08-03
fix!: enforce read authorization on referenced task lookup

ehsavoie:
Filter referenced tasks (referenceTaskIds) through TaskAuthorizationProvider before populating the RequestContext, so callers cannot read tasks they are not authorized to access.

Pass ServerCallContext to validateRequestedTask so the authorization decorator can check read permissions when streaming or subscribe requests reference an existing task. Without this, an unauthorized caller could probe for task existence via sendStreamingMessage or subscribeToTask.

- Expose getServerCallContext() on RequestContext.Builder
- Wire TaskAuthorizationProvider into SimpleRequestContextBuilder
- Inject TaskAuthorizationProvider via CDI in DefaultRequestHandler
- Propagate ServerCallContext through validateRequestedTask in all transports (JSON-RPC, REST) and decorators (OTel, auth)
- Standardize fail-closed authorization in InMemoryTaskStore to match JpaDatabaseTaskStore (deny when auth is configured but no context)
- Add unit and integration tests for reference task authorization

### DIFF
--- a/server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/AuthorizationRequestHandlerDecorator.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/AuthorizationRequestHandlerDecorator.java
@@ -261,7 +261,10 @@ public class AuthorizationRequestHandlerDecorator implements RequestHandler {
     }
 
     @Override
-    public void validateRequestedTask(@Nullable String requestedTaskId) throws A2AError {
-        delegate.validateRequestedTask(requestedTaskId);
+    public void authorizeTaskAccess(@Nullable String requestedTaskId, ServerCallContext context, TaskOperation operation) throws A2AError {
+        if (requestedTaskId != null) {
+            enforceRead(context, requestedTaskId, operation);
+        }
+        delegate.authorizeTaskAccess(requestedTaskId, context, operation);
     }
 }
--- a/server-common/src/main/resources/META-INF/a2a-defaults.properties
+++ b/server-common/src/main/resources/META-INF/a2a-defaults.properties
@@ -27,3 +27,8 @@ a2a.executor.keep-alive-seconds=60
 # Queue capacity for pending tasks (must be bounded to enable pool growth)
 # When queue is full, new threads are created up to max-pool-size
 a2a.executor.queue-capacity=100
+
+# SimpleRequestContextBuilder - Referenced task population
+# When true, referenced task IDs in messages are resolved from the TaskStore
+# and made available via RequestContext.getRelatedTasks()
+a2a.request-context.populate-referred-tasks=true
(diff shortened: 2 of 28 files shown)


server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/AuthorizationRequestHandlerDecorator.java
```diff
@@ -261,7 +261,10 @@ public void onDeleteTaskPushNotificationConfig(DeleteTaskPushNotificationConfigP
     }
 
     @Override
-    public void validateRequestedTask(@Nullable String requestedTaskId) throws A2AError {
-        delegate.validateRequestedTask(requestedTaskId);
+    public void authorizeTaskAccess(@Nullable String requestedTaskId, ServerCallContext context) throws A2AError {
+        if (requestedTaskId != null) {
+            enforceRead(context, requestedTaskId, TaskOperation.GET_TASK);
```
kabir: This seems to be called when subscribing to a task or sending a streaming message.

So GET_TASK probably isn't the right enum to use here. Not sure how best to propagate which of the methods is called.

server-common/src/main/resources/META-INF/a2a-defaults.properties
```diff
@@ -27,3 +27,8 @@ a2a.executor.keep-alive-seconds=60
 # Queue capacity for pending tasks (must be bounded to enable pool growth)
 # When queue is full, new threads are created up to max-pool-size
 a2a.executor.queue-capacity=100
+
+# SimpleRequestContextBuilder - Referenced task population
+# When true, referenced task IDs in messages are resolved from the TaskStore
+# and made available via RequestContext.getRelatedTasks()
+a2a.request-context.populate-referred-tasks=false
```
kabir: Did we discuss making it load referenced tasks by default? Not loading them seems to be a pre-existing bug

kabir (changes requested):

server-common/src/main/resources/META-INF/a2a-defaults.properties
ehsavoie: Ok, I misunderstood.

server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/AuthorizationRequestHandlerDecorator.java
ehsavoie: I think this is the better option until we have a finer grained authorization provider.

ehsavoie: I agree we could remove the operation parameter, but since it is to access task, let's leave it as it is

ehsavoie:
Should be all good @kabir

kabir (approved):

PR #1011, opened 2026-08-05
chore: Announce 1.2.0.Final release

kabir:


kabir:
Added the hold label so we don't accidentally merge before the release is done

PR #1016, opened 2026-08-06
fix(otel-it): remove dead CDI alternative and refresh stale integration-tests docs

ez-lbz:
### 1. Remove non-existent `InstrumentedRequestHandler` alternative from `beans.xml`

**Problem:** `extras/opentelemetry/integration-tests/src/main/resources/META-INF/beans.xml` declared
`org.a2aproject.sdk.extras.opentelemetry.it.InstrumentedRequestHandler` as a CDI alternative, but that class
no longer exists in the module — a leftover from an earlier test design. The beans.xml therefore declares an
alternative that cannot be resolved: invalid CDI configuration that fails deployment under strict containers
(e.g. Weld), and dead config that misleads anyone reading the module.

**Fix (beans.xml):**
- Removed the `<alternatives>` block; the file now only declares `bean-discovery-mode="all"`.

### 2. Update stale OpenTelemetry integration-tests README and make span exporter wiring explicit

**Problem:** `extras/opentelemetry/integration-tests/README.md` described components and tests that no longer
exist (`SimpleAgent`, `AgentResource`, `InstrumentedRequestHandler`, `OpenTelemetryProducer`,
`OpenTelemetryIntegrationTest`, `OpenTelemetryTracingTest`) and a "Known Issues" section claiming tests time
out on message send and that `InMemorySpanExporter` does not find expected spans — none of which reproduce
today (the module tests pass and spans are exported correctly). The span exporter was also only wired through
the Quarkus default (`cdi`), so the test harness's reliance on the in-memory exporter was implicit and easy to
break.

**Fix (A2ATestRoutes / application.properties):**
- Rewrote the README to describe the actual module: `SimpleAgentExecutor`, `A2ATestRoutes`, `TestUtilsBean`,
  `TestAgentCardProducer`, and the three test classes (`OpenTelemetryA2ATest`, `OpenTelemetryA2AIT`,
  `OpenTelemetryTest`); dropped the resolved "Known Issues".
- Set `quarkus.otel.traces.exporter=cdi` explicitly in both `src/main/resources/application.properties` and
  `src/test/resources/application.properties`, so the CDI `InMemorySpanExporter` bean is selected explicitly
  instead of by default.

### 3. Correct required Java version in helloworld example READMEs

**Problem:** `examples/helloworld/server/README.md` and `examples/helloworld/client/README.md` stated
"Java 11 or higher", but the SDK and both examples target Java 17 (parent pom sets
`maven.compiler.release=17`); users following the README with Java 11 would fail to build/run.

**Fix (helloworld READMEs):**
- Updated both READMEs to "Java 17 or higher".

## Testing

- Verified the README-described "Known Issues" do not reproduce: the module test suite was run before and
  after the changes.
- `mvn verify -pl extras/opentelemetry/integration-tests` — all 9 tests pass: `OpenTelemetryA2ATest` (4) and
  `OpenTelemetryTest` (1) in `@QuarkusTest` (JVM) mode; `OpenTelemetryA2AIT` (4) in `@QuarkusIntegrationTest`
  (packaged) mode.
- Span assertions confirm `GetTask` / `ListTasks` / `CancelTask` SERVER spans with the expected attributes
  reach `InMemorySpanExporter` via `/export`.

Behavior change: none — the fixes are config/docs only; the explicit `cdi` exporter is the same effective
behavior as the previous implicit default.


ehsavoie:
Thanks, but https://github.com/a2aproject/a2a-java/pull/1019 superseded it

PR #633, opened 2026-02-06
refactor: Update A2A protocol to align with spec revision

ehsavoie:
- Move protocol version from AgentCard to AgentInterface level
- Simplify resource identifiers from hierarchical names to flat IDs (e.g., "tasks/123" -> "123")
- Restructure Part types: remove FilePart/DataPart in favor of unified Part with url/raw/data variants
- Add OAuth flow support (ImplicitOAuthFlow, PasswordOAuthFlow)
- Rename Security to SecurityRequirement for clarity
- Standardize spelling: CANCELLED -> CANCELED
- Update all client transports (gRPC, JSON-RPC, REST) and tests
- Rename resubscribe to subscribeToTask

Breaking changes in proto API affecting:
  - Task operations (get, cancel, subscribe)
  - Push notification config management
  - Message part structure

Fixes #632 🦕

PR #1026, opened 2026-08-08
fix(grpc): transport validation and error message

### LINKED ISSUE #1014: fix: restore defensive copying of DataPart data
#633 widened `DataPart.data` from `Map<String, Object>` to `Object` and
dropped the `data = Map.copyOf(data)` defensive copy, so callers can mutate
the contents after construction. The record still copies `metadata` via
`CollectionCopies` two lines above.

Widening the type is correct — the spec allows any JSON value — but the
copy should be applied when the value is a `Map` or `List`. Also updates
the Javadoc, which currently says `data` "supports ... null" while the
constructor calls `checkNotNullParam` and the sibling constructor says
"not null".

malladinagarjuna2:
fixes #1014 
 GrpcTransport- getTaskPushNotificationConfiguration(): Replaced a manual IllegalArgumentException null check with checkNotNullParam("id", request.id()) for consistency with the validation used throughout the class.

  - subscribeToTask(): Fixed a copy-paste error in the error message prefix — it incorrectly read "Failed to subscribe task push notification config" and now reads "Failed to subscribe to task".

  DataPart immutability (regression from #633)

  PR #633 widened DataPart.data from Map<String, Object> to Object to support any JSON value (arrays, primitives), but in doing so dropped the data = Map.copyOf(data) defensive copy. As a result, callers could mutate the contents
  of a DataPart after construction.

  The widening itself is correct — the spec allows any JSON value — so the defensive copy is now applied based on the runtime type:
  - Map → CollectionCopies.unmodifiableShallowMap() (preserves null values, consistent with how metadata is already handled)
  - List → Collections.unmodifiableList(new ArrayList<>(list)) (preserves null elements)
  - Primitives / immutable values → stored as-isAlso corrected the DataPart Javadoc, which claimed data "supports null" while the constructor calls checkNotNullParam and throws on null. The null-value claim is removed from the
  class-level docs and both constructor @param tags.

  Tests- Added5 tests in DataPartTest verifying:

  - Map/List data is defensively copied and immutable (mutation throws UnsupportedOperationException, later mutation of the original input has no effect)
  - Null values in data maps and null elements in data lists survive the copy - Primitives are stored as-is (same instance)
  - Full spec module test suite passes (142 tests).


malladinagarjuna2:
@kabir  could you please verify

kabir:
@malladinagarjuna2 I opened https://github.com/malladinagarjuna2/a2a-java/pull/1 to address a few things I found. Thanks!

malladinagarjuna2:
>@malladinagarjuna2 I opened https://github.com/malladinagarjuna2/a2a-java/pull/1 to address a few things I found. Thanks!

@kabir  really thanks for the review  and sorry for the late reply, 
could you please have a look at my comments  in  https://github.com/malladinagarjuna2/a2a-java/pull/1

malladinagarjuna2:
@kabir could you please review it

kabir:
Thanks @malladinagarjuna2

PR #1028, opened 2026-08-09
feat: make AgentExecutor a functional interface with default cancel m…

### LINKED ISSUE #1009: [Feat]: Makes cancel method of AgentExecutor default

Actually AgentExecutor requires implementation of two methods, which makes it impossible to use a lambda. 

So maybe it would be great to have the cancel nethod as default (maybe throwing an exception by default) so we can always override, but if not, then we can use a functional interface.

_No response_

### Additional context

_No response_

### Code of Conduct

malladinagarjuna2:
## Overview
This PR converts the `AgentExecutor` interface into a `@FunctionalInterface` by providing a default implementation for the `cancel` method, allowing developers to define simple agents using lambda expressions.

Fixes #1009

## The Problem
Previously, `AgentExecutor` required developers to implement two methods: `execute` and `cancel`. Because there were two abstract methods, it was impossible to use a lambda expression (e.g. `(context, emitter) -> { ... }`) to implement an agent. This forced developers to write verbose boilerplate even for simple fire-and-forget agents that do not require cancellation logic.

## The Solution
1. **Added `@FunctionalInterface`**: Explicitly marked `AgentExecutor` as a functional interface.
2. **Default `cancel` Implementation**: Converted the `cancel` method to a `default` method. If an agent developer does not override it, it now correctly throws a `TaskNotCancelableError("Agent execution is not cancelable.")` by default, conforming to the existing A2A error specification.


server-common/src/main/java/org/a2aproject/sdk/server/agentexecution/AgentExecutor.java
```diff
@@ -147,5 +148,7 @@ public interface AgentExecutor {
      * @throws org.a2aproject.sdk.spec.TaskNotCancelableError if this agent does not support cancellation
      * @throws A2AError if cancellation is supported but failed to execute
      */
-    void cancel(RequestContext context, AgentEmitter emitter) throws A2AError;
+    default void cancel(RequestContext context, AgentEmitter emitter) throws A2AError {
+        throw new org.a2aproject.sdk.spec.TaskNotCancelableError("Agent execution is not cancelable.");
```
kabir: Please use an import rather than a FQN

kabir (changes requested):

server-common/src/main/java/org/a2aproject/sdk/server/agentexecution/AgentExecutor.java
malladinagarjuna2: hey sorry i didint know about the contributing guidelines well read the claude.md, yeah it is clearly mentioned about usage of imports.
will take care of this

kabir:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kabir:
Thank you @malladinagarjuna2 :-)

PR #1029, opened 2026-08-09
feat(server): allow injecting custom A2AHttpClient for push notificat…

### LINKED ISSUE #135: [Feat]: Make it possible to plug in other A2AHttpClient implements for the server side

_No response_

https://github.com/a2aproject/a2a-java/pull/129 pulled out server related code from core into a separate server-common module. With this change, `InMemoryPushNotifier` was modified to use the `JdkA2AHttpClient`. We should make sure it's possible to use a different `A2AHttpClient` on the server side.

_No response_

### Additional context

_No response_

### Code of Conduct

malladinagarjuna2:
## Overview
This PR introduces the ability to plug in a custom `A2AHttpClient` implementation on the server side by natively supporting CDI injection in the `BasePushNotificationSender`.

Fixes #135

## The Problem
When server-related code was extracted from `core` into `server-common` (in #129), the Push Notification sender was modified to hardcode its usage of `JdkA2AHttpClient` via `A2AHttpClientFactory.create()`. This prevented developers from replacing the underlying HTTP client on the server side (e.g. replacing the JDK client with the Vert.x HTTP client, or providing a custom client with specific interceptors/auth mechanisms).

## The Solution
1. **Added `Instance<A2AHttpClient>` to CDI Constructor:** The `@Inject` constructor in `BasePushNotificationSender` now accepts an `Instance<A2AHttpClient>`. 
2. **Graceful Fallback:** If the DI framework (like Quarkus or Spring) has a custom `A2AHttpClient` bean available (`!httpClientInstance.isUnsatisfied()`), it will use it. Otherwise, it gracefully falls back to `A2AHttpClientFactory.create()`, ensuring 100% backwards compatibility for users who don't provide a custom bean.


server-common/src/main/java/org/a2aproject/sdk/server/tasks/BasePushNotificationSender.java
```diff
@@ -66,8 +66,13 @@ public BasePushNotificationSender(PushNotificationConfigStore configStore) {
 
     @Inject
     public BasePushNotificationSender(PushNotificationConfigStore configStore,
-                                       Instance<PushNotificationPayloadFormatter> formatters) {
-        this.httpClient = A2AHttpClientFactory.create();
+                                       Instance<PushNotificationPayloadFormatter> formatters,
+                                       Instance<A2AHttpClient> httpClientInstance) {
+        if (!httpClientInstance.isUnsatisfied()) {
```
kabir: Please use `if (httpClientInstance.isResolvable()` instead. It is the project convention, and helps guard against several beans being found

kabir (changes requested):

kabir:
Also, we could do with some tests for the new code path :-)

server-common/src/main/java/org/a2aproject/sdk/server/tasks/BasePushNotificationSender.java
malladinagarjuna2: got it!!

ehsavoie:
Why not providing an A2AHttpClientProvider implementation using CDI.current() to get your CDI httpclient ?

ehsavoie:
This is what I meant: https://github.com/a2aproject/a2a-java/pull/1071 Would that fix your issue @malladinagarjuna2 ?

malladinagarjuna2:
>Why not providing an A2AHttpClientProvider implementation using CDI.current() to get your CDI httpclient ?

hey i am sorry for  the late reply @ehsavoie  
>This is what I meant: https://github.com/a2aproject/a2a-java/pull/1071 Would that fix your issue @malladinagarjuna2 >?

Yes, that fixes it, thanks for creating the pr, it is the better approach: well doing it at the A2AHttpClientFactory level means every create() call site benefits, not just BasePushNotificationSender, and it keeps server-common untouched. I'll close the curent pr in favour of it. 

well actually after your message 
>Why not providing an A2AHttpClientProvider implementation using CDI.current() to get your CDI httpclient ?
i was about to do that  fix 


One small thing worth which is important since create() now eats  provider failures at fine level, a user who intends to use their CDI client but has a misconfigured bean will silently get the JDK client instead. Might be worth logging that at WARNING when a provider that reported a high priority fails, so it isn't invisible.

I also have a Weld SE test locally that boots a real container with a produced A2AHttpClient bean and asserts A2AHttpClientFactory.create() returns it end-to-end well it  complements the mockStatic(CDI.class) unit tests with a real-container check. Happy to open it as a PR against your branch if useful.

ehsavoie:
Superseded by https://github.com/a2aproject/a2a-java/pull/1071

PR #1030, opened 2026-08-09
docs: add multi-agent reference architecture

### LINKED ISSUE #898: How to scale a A2A production deployment to 100 A2A Server Agents each exposing a different AgentCard?
Say i have an quarkus App that implemented A2A SDK with on AgentExecutor and one AgentCard and i would like to scale this deployment to support a higher number of Server Agents.

the one approach that would work would be to deploy one separate Quarkus app per Agent to kubernetes and proxy the connection with a reverse proxy software creating virtual hosts/ingresses to access the Server Agent endpoint (AgentCard and JSONRPC handler)

```
https://a2a-server-agent-1/.well-known/agent-card.json
https://a2a-server-agent-2/.well-known/agent-card.json
etc
```
I think that's a lot overhead running a separate quarkus app per agent

is there any best practices in terms of setting up just one single quarkus app supporting multiple AgentCards and JSONRPC Handler endpoints?

is there any reference architecture for a typical A2A production deployment?

malladinagarjuna2:
## Overview
This PR adds a reference architecture documentation page detailing how to scale A2A Server Agents by hosting multiple distinct agents within a single Quarkus application.

Fixes #898

## The Problem
Deploying a single agent per Quarkus app (1:1 mapping to kubernetes pods) incurs significant operational overhead and resource consumption when the number of agents grows into the hundreds. Users needed official guidance and a reference architecture for deploying multiple `AgentCard`s and JSON-RPC handlers from a single application platform.

## The Solution
Added a dedicated `Multi-Agent Scaling` page to the documentation site. The guide explains how to bypass the default `@ApplicationScoped` singletons and manually register Vert.x routing paths for individual agents (e.g., `/{agentId}/.well-known/agent-card.json`), allowing a single app to dynamically host an unlimited number of agents with minimal overhead.


kabir (changes requested):
Thanks for the PR :-)

Before we merge, I think we can take a better approach. Rather than documenting the low-level Vert.x wiring as pseudocode in a docs page, it could perhaps be more useful to add an actual utility class/method to the SDK that handles multi-agent registration. Then the docs page just shows how to use it.

A few reasons:

1. The current code example is incomplete — the JSON-RPC handler route body is // ..., and manually constructing a JSONRPCHandler outside CDI is non-trivial. A utility would encapsulate that complexity.
2. It's JSON-RPC-only — the same pattern applies to gRPC and REST transports. A utility can abstract across all three.
3. It's testable — unlike documentation pseudocode, a utility class can have tests to make sure it actually works.
4. Title inconsistency — the page says "Multi-Agent Deployments" but the menu says "Multi-Agent Scaling" (minor, but worth aligning).

Also, @mescaja shared a complementary approach on https://github.com/a2aproject/a2a-java/issues/898#issuecomment-5240570449 using
  database-backed agent cards and @PathParam. I've not looked enough to see if that is also something that is possible to pull into this utility. 

The ideal outcome would be:
- A utility in the SDK that makes multi-agent setup easy
- A short docs page showing how to use it

What do you think?

kabir:
Actually https://github.com/a2aproject/a2a-java/pull/1031 is probably the utility :-) https://github.com/a2aproject/a2a-java/pull/1031. I think these two PRs would be better put together into one so we get both docs and implementation in one go

malladinagarjuna2:
alright kabir will close this pr  and include  the doc files into #1031

malladinagarjuna2:
@kabir  added the docs in #1031

PR #1045, opened 2026-08-10
fix: enforce task state-machine transitions and serialize cancels

ez-lbz:
### 1. Enforce state-machine transitions in `TaskManager` 

**Problem:** `TaskManager.saveTaskEvent`/`process` overwrote the persisted task status with whatever state the event carried, with no transition validation. A task in a terminal state (`COMPLETED`/`FAILED`/`CANCELED`/`REJECTED`) could be silently rewritten to a different state (e.g. `COMPLETED` → `SUBMITTED`) by a late, stale, or malformed event — including replicated events racing the local final event. Only Go partially blocks this today.

**Fix (server-common/src/main/java/org/a2aproject/sdk/server/tasks/TaskManager.java):**
- Added `validateStateTransition(currentState, newState, taskId)` and call it in the status-update path (`saveTaskEvent(TaskStatusUpdateEvent)`) and the full-task path (`saveTaskEvent(Task)`).
- Rule: a terminal state must not be overwritten by a **different** state (throws `A2AServerException`, which the event pipeline turns into an error to the client while preserving the persisted state). Re-arriving events carrying the **same** final state remain allowed, so replicated replays and idempotent retries keep working.
- All transitions from non-terminal states stay permitted (SUBMITTED → WORKING → COMPLETED/FAILED/CANCELED, interrupted-state resume flows INPUT_REQUIRED/AUTH_REQUIRED → WORKING → COMPLETED), so normal `AgentExecutor` flows are unaffected.

**Fix (server-common/src/test/java/org/a2aproject/sdk/server/tasks/TaskManagerTest.java):**
- `testRejectStatusUpdateOverwritingTerminalState`, `testRejectStatusUpdateToDifferentTerminalState`, `testRejectTaskEventOverwritingTerminalState` — rejected transitions throw and the persisted terminal state is preserved.
- `testSameTerminalStateReplayAllowed` — idempotent same-state replay still works.
- `testNormalStateFlowAllowed`, `testInterruptedStateResumeFlowAllowed` — the standard flows keep working.

### 2. Serialize concurrent cancels per task 

**Problem:** `DefaultRequestHandler.onCancelTask` performed a check-then-act sequence — read task → check `isFinal()` → invoke `agentExecutor.cancel()` — with no lock between the check and the act. Two concurrent cancels of the same task could both observe the pre-transition state and both "succeed", and a cancel could race a concurrent completion.

**Fix (server-common/src/main/java/org/a2aproject/sdk/server/requesthandlers/DefaultRequestHandler.java):**
- Added a per-task lock registry (`cancelLocks`, a `ConcurrentHashMap<String, Object>` keyed by task ID) and moved the entire cancel body into `synchronized (lock)` via a `doCancelTask` helper. The second concurrent cancel now waits for the first to finish, observes the `CANCELED` terminal state, and fails with `TaskNotCancelableError`.
- The state-machine validation additionally blocks a cancel from overwriting a task that a concurrent `message/send` completed first.

**Fix (server-common/src/test/java/org/a2aproject/sdk/server/requesthandlers/DefaultRequestHandlerTest.java):**
- `testConcurrentCancelsAreSerialized` — holds the first cancel inside `agentExecutor.cancel()`, asserts the second cancel blocks, then verifies the first succeeds with `CANCELED` and the second fails with `TaskNotCancelableError`.

Behavior change: (1) events attempting to change a terminal task's state are now rejected instead of silently overwriting the state; (2) concurrent cancels of the same task are serialized, so the second one gets `TaskNotCancelableError` instead of both succeeding.

### 2. Make the replicated queue manager parallel test deterministic

**Problem:** `ReplicatedQueueManagerTest.testParallelReplicationBehavior` was timing-dependent and failed intermittently in CI (observed counts 1, 2, 21 instead of the expected 25) and consistently locally (0 or 3). The replicated threads sent `TASK_STATE_COMPLETED` events; a COMPLETED event processed mid-stream finalizes the task and closes the queue, so overlapping normal enqueues no longer trigger replication and the final count assertion depends on thread interleaving.

**Fix (extras/queue-manager-replicated/core/src/test/java/org/a2aproject/sdk/extras/queuemanager/replicated/core/ReplicatedQueueManagerTest.java):**
- Replicated threads now send a non-terminal state (`TASK_STATE_WORKING`). The replication hook skips replicated events via `isReplicated()` regardless of state, so the test's intent (normal enqueues replicate, replicated events do not) is unchanged while the outcome is deterministic.

## Testing

- `mvn -pl extras/queue-manager-replicated/core test -Dtest=ReplicatedQueueManagerTest` — **15 tests run, 0 failures** across 5 consecutive runs (previously failed 3/3 locally with the same command).

### DIFF
--- a/server-common/src/main/java/org/a2aproject/sdk/server/tasks/TaskManager.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/tasks/TaskManager.java
@@ -19,6 +19,7 @@ import org.a2aproject.sdk.spec.InternalError;
 import org.a2aproject.sdk.spec.Message;
 import org.a2aproject.sdk.spec.Task;
 import org.a2aproject.sdk.spec.TaskArtifactUpdateEvent;
+import org.a2aproject.sdk.spec.TaskState;
 import org.a2aproject.sdk.spec.TaskStatus;
 import org.a2aproject.sdk.spec.TaskStatusUpdateEvent;
 import org.jspecify.annotations.Nullable;
@@ -69,6 +70,13 @@ public class TaskManager {
     boolean saveTaskEvent(Task task, boolean isReplicated, @Nullable AtomicReference<Task> taskSnapshot)
             throws A2AServerException {
         checkIdsAndUpdateIfNecessary(task.id(), task.contextId());
+        // Defensive state-machine check: a task that already reached a terminal state must
+        // not be overwritten by a task snapshot carrying a different state (BUG-43).
+        Task current = getTask();
+        if (current != null && current.status() != null && current.status().state() != null
+                && task.status() != null && task.status().state() != null) {
+            validateStateTransition(current.status().state(), task.status().state(), task.id());
+        }
         Task savedTask = saveTask(task, isReplicated);
         if (taskSnapshot != null) {
             taskSnapshot.set(savedTask);
@@ -85,6 +93,12 @@ public class TaskManager {
         checkIdsAndUpdateIfNecessary(event.taskId(), event.contextId());
         Task task = ensureTask(event.taskId(), event.contextId());
 
+        // State-machine validation: reject transitions that would overwrite a terminal
+        // state with a different state (BUG-43). Re-arriving events carrying the same
+        // final state remain allowed (idempotent replays / replication).
+        TaskState currentState = task.status() != null ? task.status().state() : null;
+        TaskState newState = event.status() != null ? event.status().state() : null;
+        validateStateTransition(currentState, newState, event.taskId());
 
         Task.Builder builder = Task.builder(task)
                 .status(event.status());
@@ -230,6 +244,36 @@ public class TaskManager {
             saveTask(task, false);  // Local operation, not replicated
         }
         return task;
+    }
+
+    /**
+     * Validates a task state transition before it is persisted (BUG-43).
+     * <p>
+     * A terminal (final) state must not be overwritten by a <em>different</em> state:
+     * once a task is {@code COMPLETED}/{@code FAILED}/{@code CANCELED}/{@code REJECTED}
+     * it stays in that state. Events re-arriving with the <em>same</em> final state are
+     * allowed, so replicated replays and idempotent retries keep working.
+     * <p>
+     * Transitions from any non-terminal state to any state are permitted (e.g.
+     * SUBMITTED → WORKING → COMPLETED/FAILED/CANCELED, interrupted-state resume flows),
+     * matching the transitions the A2A spec and the reference agents exercise.
+     *
+     * @param currentState the task's current state, or {@code null} if unknown
+     * @param newState     the state requested by the event, or {@code null} if unknown
+     * @param taskId       the task identifier, used in the error message
+     * @throws A2AServerException if the transition would overwrite a terminal state
+     */
+    private static void validateStateTransition(@Nullable TaskState currentState, @Nullable TaskState newState,
+            String taskId) throws A2AServerException {
+        if (currentState == null || newState == null) {
+            return;
+        }
+        if (currentState.isFinal() && currentState != newState) {
+            throw new A2AServerException(
+                    "Task " + taskId + " is already in terminal state " + currentState
+                            + " and cannot transition to " + newState,
+                    new InternalError("Task " + taskId + " is already in terminal state " + currentState));
+        }
     }
 
     private Task createTask(String taskId, String contextId) {
--- a/server-common/src/main/java/org/a2aproject/sdk/server/tasks/TaskManager.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/tasks/TaskManager.java
@@ -182,13 +182,28 @@ public class TaskManager {
 
             // Only create status update if we have contextId
             if (errorContextId != null) {
-                LOGGER.debug("A2AError event detected, transitioning task {} to FAILED", taskId);
-                TaskStatusUpdateEvent failedEvent = TaskStatusUpdateEvent.builder()
-                        .taskId(taskId)
-                        .contextId(errorContextId)
-                        .status(new TaskStatus(TASK_STATE_FAILED))
-                        .build();
-                isFinal = saveTaskEvent(failedEvent, isReplicated, taskSnapshot);
+                // If the task is already in a terminal state, skip the state
+                // update entirely: the synthesized FAILED event would be rejected
+                // by the state-machine guard (terminal -> FAILED is a terminal-to-
+                // different-terminal transition), and the resulting exception would
+                // surface to clients as a misleading internal error. The A2AError
+                // itself still signals finality to clients.
+                Task existingTask = getTask();
+                TaskState currentState = existingTask != null && existingTask.status() != null
+                        ? existingTask.status().state() : null;
+                if (currentState != null && currentState.isFinal()) {
+                    LOGGER.debug("A2AError event for task {} already in terminal state {} - skipping state update",
+                            taskId, currentState);
+                    isFinal = true;
+                } else {
+                    LOGGER.debug("A2AError event detected, transitioning task {} to FAILED", taskId);
+                    TaskStatusUpdateEvent failedEvent = TaskStatusUpdateEvent.builder()
+                            .taskId(taskId)
+                            .contextId(errorContextId)
+                            .status(new TaskStatus(TASK_STATE_FAILED))
+                            .build();
+                    isFinal = saveTaskEvent(failedEvent, isReplicated, taskSnapshot);
+                }
             } else {
                 // Can't update status without contextId, but error is still terminal
                 LOGGER.debug("A2AError event for task {} without contextId - skipping state update", taskId);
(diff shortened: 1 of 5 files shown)


server-common/src/test/java/org/a2aproject/sdk/server/tasks/TaskManagerTest.java
```diff
@@ -739,4 +739,123 @@ public void testUpdateWithMessage() throws A2AServerException {
...
+        Task completedTask = Task.builder()
+                .id("task-terminal")
+                .contextId("ctx-1")
+                .status(new TaskStatus(TaskState.TASK_STATE_COMPLETED))
+                .build();
+        taskStore.save(completedTask, false);
+        TaskManager tm = new TaskManager("task-terminal", "ctx-1", taskStore, null);
+
+        // A status update to a different state after the terminal state must be rejected
+        TaskStatusUpdateEvent workingEvent = TaskStatusUpdateEvent.builder()
```
ehsavoie, line 756: Some helper method would help reduce the duplication code on creating those

server-common/src/main/java/org/a2aproject/sdk/server/tasks/TaskManager.java
ehsavoie, line 140: I think this should be guarded as well: if the task is already COMPLETED then it go wild

ehsavoie, line 186: This would set the Task to TASK_STATE_FAILED, but this could lead to a new Error because of the guard if the task is in any terminal state

ehsavoie (changes requested):
Trace the full call chain:

  TaskManager.java:161-191 — process() handles A2AError by synthesizing a TASK_STATE_FAILED event and calling saveTaskEvent:
  TaskStatusUpdateEvent failedEvent = ...status(new TaskStatus(TASK_STATE_FAILED)).build();
  isFinal = saveTaskEvent(failedEvent, isReplicated, taskSnapshot);  // line 191

  TaskManager.java:101 — saveTaskEvent(TaskStatusUpdateEvent,...) calls the new guard:
  validateStateTransition(currentState, newState, event.taskId());

  If the task is already COMPLETED (or any other terminal state), validateStateTransition at line 266-276 throws A2AServerException. Before this PR, it would silently overwrite COMPLETED with FAILED.

  MainEventBusProcessor.java:343-347 — updateTaskStore() catches this as the generic catch (Exception e) block (because A2AServerException is not InternalError, TaskSerializationException, or TaskPersistenceException):
  } catch (Exception e) {
      // Unexpected exception type - treat as permanent failure
      throw new InternalError("TaskStore persistence failed: " + e.getMessage());
  }

  MainEventBusProcessor.java:230-234 — back in processEvent(), that InternalError is caught and set as the event to distribute to clients:
  } catch (InternalError e) {
      LOGGER.error("Failed to persist event for task {}, distributing error to clients", taskId, e);
      eventToDistribute = e;  // clients receive this InternalError instead of the original A2AError

  Net effect: a client that sent a message to an already-completed task, whose agent then emits an A2AError, now receives a generic InternalError("TaskStore persistence failed: Task X is already in terminal state COMPLETED") instead of the original A2AError. The message is confusing
  because the store didn't fail — the state machine rejected the transition.

  The cleaner fix would be inside TaskManager.process() itself at line 161: before synthesizing the FAILED event, check if the task is already in a terminal state and skip the update (just return true):
  } else if (event instanceof A2AError) {
      // ... existing null checks ...
      if (errorContextId != null) {
          Task existing = getTask();
          TaskState currentState = existing != null && existing.status() != null ? existing.status().state() : null;
          if (currentState != null && currentState.isFinal()) {
              // Task already terminal — no state update needed, A2AError still signals finality
              return true;
          }   
          // ... synthesize FAILED event as before

  This would prevent the A2AServerException from propagating at all, and clients would still receive the original A2AError (since process() returns normally, eventToDistribute stays as the original event in processEvent()).

ez-lbz:
Thanks for the detailed call-chain trace — that's a real regression from the state-machine guard. Fixed as you suggested: in `TaskManager.process()`, the `A2AError` branch now checks whether the task is already terminal before synthesizing the `TASK_STATE_FAILED` event; if so, it skips the state update and lets the `A2AError` signal finality (returns `true`). The exception no longer propagates, so clients receive the original `A2AError` instead of the misleading "TaskStore persistence failed" `InternalError`.

Added regression tests for both paths (terminal task + non-terminal task). `TaskManagerTest` 36 passed.

ehsavoie (changes requested):
1. Unbounded cancelLocks map (Medium)

  The problem: cancelLocks at DefaultRequestHandler.java:270 grows monotonically — computeIfAbsent adds an entry per canceled task, but there's no corresponding remove anywhere. By contrast, runningAgents (line 258) is explicitly cleaned up at lines 744 and 908.

  Why it matters practically: The Javadoc (line 267) argues it's "bounded by the number of distinct tasks that have ever been canceled, in the same way the in-memory task store is unbounded." That's true today with InMemoryTaskStore, but with the JPA-backed task-store-database-jpa
  extra, the task store is not unbounded in memory — tasks live in the DB. In that configuration, cancelLocks becomes the only in-memory structure that grows without bound.

  The fix is straightforward. After doCancelTask returns (line 485) — whether successfully or via TaskNotCancelableError — the task is guaranteed to be in a terminal state. No future cancel can succeed (the state-machine guard in TaskManager.validateStateTransition will reject it
  even without the lock). So the lock entry is no longer needed:

  synchronized (cancelLock) {
      try {
          return doCancelTask(params, context);
      } finally {
          cancelLocks.remove(params.id());
      }
  }

  The finally ensures cleanup on both the success path and when doCancelTask throws TaskNotCancelableError (task was already terminal). A brief window exists where a third concurrent cancel could create a new lock object after the remove, but that's safe — it will just immediately
  see the terminal state and throw.

  ---
  2. Duplicate terminal-state guard logic (Medium)
  
  The three locations:

  1. TaskManager.java:78 — saveTaskEvent(Task, ...) calls validateStateTransition
  2. TaskManager.java:101 — saveTaskEvent(TaskStatusUpdateEvent, ...) calls validateStateTransition
  3. TaskManager.java:191-197 — process(A2AError, ...) has its own inline check:
  Task existingTask = getTask();
  TaskState currentState = existingTask != null && existingTask.status() != null
          ? existingTask.status().state() : null;
  if (currentState != null && currentState.isFinal()) { ... }

  Why they diverge: Locations 1 and 2 reject the transition by throwing. Location 3 silently skips it and sets isFinal = true. The skip is intentional — an A2AError arriving for a terminal task should not throw, because the error itself still needs to propagate to clients. But the
  terminal-state detection logic (the null-check chain + isFinal() call) is duplicated.

  The risk: If the definition of "terminal" ever changes (e.g., a new TASK_STATE_ARCHIVED that is final but allows certain transitions), you'd need to update both validateStateTransition and the inline check. They're in the same file, but the inline check at line 191 doesn't call
  validateStateTransition or even reference it, so the connection is invisible.

  A possible consolidation: Extract a isInTerminalState() query method (or just use getTask() + isFinal() consistently) and restructure process() to attempt the saveTaskEvent and catch the A2AServerException:

  if (errorContextId != null) {
      TaskStatusUpdateEvent failedEvent = TaskStatusUpdateEvent.builder()
              .taskId(taskId).contextId(errorContextId)
              .status(new TaskStatus(TASK_STATE_FAILED)).build();
      try {
          isFinal = saveTaskEvent(failedEvent, isReplicated, taskSnapshot);
      } catch (A2AServerException e) {
          // Task already terminal — the state-machine guard rejected FAILED.
          // The A2AError still signals finality to clients.
          LOGGER.debug("A2AError for task {} already in terminal state — skipping state update", taskId);
          isFinal = true;
      }
  }

  This eliminates the duplicate check entirely — validateStateTransition inside saveTaskEvent is the single authority. The downside is using exception flow for a non-exceptional path, but this is a rare edge case (error arriving after completion), not a hot path.

ez-lbz:
Both points addressed:

1. **Unbounded cancelLocks**: `onCancelTask` now drops the per-task lock entry in a `finally` after `doCancelTask` returns (success or `TaskNotCancelableError`), since the task is then terminal and the state-machine guard rejects any future cancel without the lock. The 2-arg `remove` keeps a concurrently-created newer entry intact.

2. **Duplicated terminal-state detection**: reworked the `A2AError` branch to attempt the synthesized `TASK_STATE_FAILED` transition and catch the `A2AServerException` from `validateStateTransition` — the single authority — instead of the inline pre-check. The A2AError still signals finality to clients.

`TaskManagerTest` + `DefaultRequestHandlerTest` — 58 tests, 0 failures.

ehsavoie:
@ez-lbz it looks like your commits are not signed https://github.com/a2aproject/a2a-java/pull/1045/checks?check_run_id=94136356379

ehsavoie (approved):

ez-lbz:
Thanks for the approval! Regarding the sign-off note — acknowledged. The commits on this branch were authored before we standardized on a Signed-off-by workflow, so they don't carry the DCO trailer yet. Happy to add `Signed-off-by` to all commits on the branch if you'd like it before merging — just say the word and I'll rebase with `-s` and push.

PR #1059, opened 2026-08-12
fix(grpc): align unsupported operation status

### LINKED ISSUE #1051: fix: UnsupportedOperationError gRPC status should be FAILED_PRECONDITION per spec
## Problem

The A2A specification (section 5.4) defines the following gRPC status mapping for `UnsupportedOperationError`:

| A2A Error Type | JSON-RPC Code | gRPC Status | HTTP Status |
|:---|:---|:---|:---|
| `UnsupportedOperationError` | `-32004` | `FAILED_PRECONDITION` | `400 Bad Request` |

However, `A2AErrorCodes.UNSUPPORTED_OPERATION` in the SDK maps to `UNIMPLEMENTED`:

https://github.com/a2aproject/a2a-java/blob/643f1f803e992892b1bac89c8ccfb9481e563141/spec/src/main/java/org/a2aproject/sdk/spec/A2AErrorCodes.java#L29

```java
UNSUPPORTED_OPERATION(-32004, "UNIMPLEMENTED", 400),
```

This diverges from the spec table, which says the gRPC status should be `FAILED_PRECONDITION`.

## Expected behavior

`UNSUPPORTED_OPERATION` should map to gRPC `FAILED_PRECONDITION` to match the spec.

## Context

Found during review of #1041, which correctly changed streaming-not-supported errors from `InvalidRequestError` to `UnsupportedOperationError`. The gRPC status mismatch is a pre-existing issue unrelated to that PR.

dangzitou:
# Description

Align `UnsupportedOperationError` with the A2A gRPC specification by mapping it to `FAILED_PRECONDITION` instead of `UNIMPLEMENTED`. Update the existing server and client mapping assertions, plus the handler mapping documentation.

- [x] Follow the [`CONTRIBUTING` Guide](../blob/main/CONTRIBUTING.md).
- [x] Use a Conventional Commits PR title.
- [x] Ensure the affected tests pass.
- [x] Appropriate documentation was updated.

## Reproduction

On current `main`, changing only the eight existing `GrpcHandlerTest` expectations for paths that emit `UnsupportedOperationError` produced 8 consistent failures out of 42 tests:

```
expected: <FAILED_PRECONDITION> but was: <UNIMPLEMENTED>
```

## Verification

Run with OpenJDK 17:

```bash
mvn -DskipTests install -pl transport/grpc,client/transport/grpc -am
mvn -pl transport/grpc test
mvn -pl client/transport/grpc test
```

Results:

- `transport/grpc`: 42 tests passed
- `client/transport/grpc`: 9 tests passed
- all 12 prerequisite/affected reactor modules compiled and installed successfully

A root `mvn clean install` was also attempted, but the unrelated `client/transport/jsonrpc` tests could not bind/use their hard-coded local endpoint because `127.0.0.1:4001` was already occupied by another local application. The two changed gRPC modules were fully tested separately above.

This fixes #1051

### DIFF
--- a/client/transport/grpc/src/main/java/org/a2aproject/sdk/client/transport/grpc/GrpcErrorMapper.java
+++ b/client/transport/grpc/src/main/java/org/a2aproject/sdk/client/transport/grpc/GrpcErrorMapper.java
@@ -74,7 +74,8 @@ public class GrpcErrorMapper {
         String desc = message != null ? message : e.getMessage() == null ? "" : e.getMessage();
         return switch (code) {
             case NOT_FOUND -> new A2AClientException(errorPrefix + desc, new TaskNotFoundError());
-            case UNIMPLEMENTED -> new A2AClientException(errorPrefix + desc, new UnsupportedOperationError());
+            case UNIMPLEMENTED, FAILED_PRECONDITION ->
+                    new A2AClientException(errorPrefix + desc, new UnsupportedOperationError());
             case INVALID_ARGUMENT -> new A2AClientException(errorPrefix + desc, new InvalidParamsError());
             case INTERNAL -> new A2AClientException(errorPrefix + desc, new org.a2aproject.sdk.spec.InternalError(null, desc, null));
             case UNAUTHENTICATED -> new A2AClientException(errorPrefix + A2AErrorMessages.AUTHENTICATION_FAILED);
--- a/client/transport/grpc/src/test/java/org/a2aproject/sdk/client/transport/grpc/GrpcErrorMapperTest.java
+++ b/client/transport/grpc/src/test/java/org/a2aproject/sdk/client/transport/grpc/GrpcErrorMapperTest.java
@@ -119,6 +119,20 @@ public class GrpcErrorMapperTest {
     }
 
     @Test
+    public void testUnsupportedOperationErrorFallbackUsesFailedPrecondition() {
+        String errorMessage = "Operation not supported";
+        StatusRuntimeException grpcException = Status.FAILED_PRECONDITION
+                .withDescription(errorMessage)
+                .asRuntimeException();
+
+        A2AClientException result = GrpcErrorMapper.mapGrpcError(grpcException);
+
+        assertNotNull(result);
+        assertNotNull(result.getCause());
+        assertInstanceOf(UnsupportedOperationError.class, result.getCause());
+    }
+
+    @Test
     public void testInvalidParamsErrorUnmarshalling() {
         String errorMessage = "Invalid parameters provided";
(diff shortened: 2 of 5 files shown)


kabir (changes requested):
Hi @dangzitou - thank you for the PR!

One small issue:
The fallback switch in GrpcErrorMapper.java (line 77) still maps UNIMPLEMENTED to UnsupportedOperationError. Since the server now sends FAILED_PRECONDITION for this error, this fallback won't match when ErrorInfo details are absent. Consider adding a
  FAILED_PRECONDITION case alongside the existing UNIMPLEMENTED one (since UNIMPLEMENTED is still used by other error types like PushNotificationNotSupportedError).

  The primary ErrorInfo-based path (lines 62-70) handles this correctly via the reason field, so this only affects the fallback.

  https://github.com/a2aproject/a2a-java/blob/2fc099dfbd19b0b82edfb6811eefaee3bff10ebe/client/transport/grpc/src/main/java/org/a2aproject/sdk/client/transport/grpc/GrpcErrorMapper.java#L73-L84

dangzitou:
Addressed the requested fallback mapping in 2d1a7cba: `FAILED_PRECONDITION` now maps to `UnsupportedOperationError` alongside `UNIMPLEMENTED`, with a no-`ErrorInfo` regression test. Rebased onto current `main`; local `GrpcErrorMapperTest` passes.

kabir:
Hi, @dangzitou sorry for not noticing this earlier but after looking a bit more carefully I noticed this isn't quite right.

The current SDK targets v1.0.0, which is still the latest released version on the official A2A spec site (https://a2a-protocol.org/latest/specification/). (source at https://github.com/a2aproject/A2A/blob/v1.0.0/docs/specification.md#54-error-code-mappings).

The  v1.0.0 spec maps UnsupportedOperationError to UNIMPLEMENTED, so I think the existing code is correct.

The FAILED_PRECONDITION mapping comes from v1.0.1 (tagged on GitHub but not yet reflected on the official site). We haven't decided whether to adopt v1.0.1  yet, and if we do, the error mapping table has 6 changes between the two versions — not just this one — so we'd want to handle them together.

I was investigating whether we could support both 1.0.0 and 1.0.1 of the protocol but both seem to use `1.0` as the version string so there is no way to distinguish between the two cases.
* https://github.com/a2aproject/A2A/blob/v1.0.0/docs/specification.md#36-versioning
* https://github.com/a2aproject/A2A/blob/v1.0.1/docs/specification.md#36-versioning

Since https://a2a-protocol.org/latest/specification/ still points to 1.0.0 I think we need to go with what we have.

Thank you for your work, but we will close this for now.

PR #1060, opened 2026-08-12
fix(server-common): keep task streams open after messages

### LINKED ISSUE #1037: agentEmitter.requiresInput(a2aMessage) does not trigger a TaskStatusUpdateEvent
I have come across an issue when setting the A2A Agent task to TaskState.TASK_STATE_INPUT_REQUIRED

These are the steps to reproduce with latest 1.2.0.Final version

**Step 1:** A2A client agent sends a a2aMessage1 with null contextId and null taskId
**Step 2**: A2A Server agent

```
agentEmitter.submit();
agentEmitter.startWork();
agentEmitter.updateStatus(TaskState.TASK_STATE_WORKING, null);

//Do some stuff, call APIs and MCPs endpoint to run tools feeding back the result
.....

//Send a message
agentEmitter.sendMessage(a2aMessage2);

//Do some more stuff, workout logic to see if the task status needs changing
....

// Change task status
agentEmitter.requiresInput(a2aMessage3);

//Do some more stuff, send an artifact and workout logic to see if the task needs to be completed
....
agentEmitter.addArtifact(List.of(textPart));
agentEmitter.complete(a2aMessage4);
```

**Step 3**: A2A Client agent

- Received MessageEvent and a2aMessage2 successfully returned using messageEvent.getMessage()
- taskStatusUpdateEvent was not received for the task status update to `TaskState.TASK_STATE_INPUT_REQUIRED` hence a2aMessage3 got lost.
- Received TaskArtifactUpdateEvent along with the TextPart using taskArtifactUpdateEvent.artifact().parts()
- Received taskStatusUpdateEvent and a2aMessage4 successfully returned using taskStatusUpdateEvent.status().message()

I tried setting a task status update using the code below. The outcome was the same

`agentEmitter.updateStatus(TaskState.TASK_STATE_INPUT_REQUIRED, a2aMessage3);`

i need to be able to send a few messages and for the A2A client to be able to stream the corresponing MessageEvent and also to be able to set the task status to TASK_STATE_INPUT_REQUIRED and stream the associated message with it afterwards.

For some reason setting a task to INPUT_REQUIRED does not trigger the  TaskStatusUpdateEvent on the client side.

I think the combination of agentEmitter.sendMessage follow by an agentEmitter.requiresInput is what causes this issue.

dangzitou:
# Description

Keep a task stream open when an agent emits a `Message` after the task lifecycle has started. A stateless `Message` remains a complete response and still closes its stream.

The new regression test covers `WORKING Task -> Message -> INPUT_REQUIRED -> COMPLETED`, including a custom message without task/context IDs as accepted by `AgentEmitter.sendMessage(Message)`.

- [x] Follow the [`CONTRIBUTING` Guide](../blob/main/CONTRIBUTING.md).
- [x] Use a Conventional Commits PR title.
- [x] Ensure the affected tests pass.
- [x] Appropriate documentation was updated.

## Reproduction

On current `main`, the focused Java 17 regression was run three times. Every run failed after delivering only the working task and message:

```
expected: <4> but was: <2>
```

The following `INPUT_REQUIRED` and `COMPLETED` updates were lost because every `Message` was treated as stream-terminal.

## Verification

Run with OpenJDK 17:

```bash
mvn -pl server-common -Dtest=EventConsumerTest test
mvn -pl server-common test
```

Results:

- `EventConsumerTest`: 19 tests passed
- `server-common`: 475 tests passed
- existing stateless-message stream termination coverage remains green

This fixes #1037

### DIFF
--- a/server-common/src/main/java/org/a2aproject/sdk/server/events/EventConsumer.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/events/EventConsumer.java
@@ -228,9 +228,9 @@ public class EventConsumer {
                             boolean isFinalEvent = false;
                             if (event instanceof TaskStatusUpdateEvent tue && tue.isFinal()) {
                                 isFinalEvent = true;
-                            } else if (event instanceof Message) {
-                                // Per A2A spec §3.1.2 (Send Streaming Message): a Message is the
-                                // complete response — the stream must close after delivering it.
+                            } else if (event instanceof Message && lastSeenTaskState == null) {
+                                // A stateless Message is the complete response. Messages emitted
+                                // after a task has started are intermediate task stream events.
                                 isFinalEvent = true;
                             } else if (event instanceof Task task) {
                                 isFinalEvent = isStreamTerminatingTask(task);
--- a/server-common/src/test/java/org/a2aproject/sdk/server/events/EventConsumerTest.java
+++ b/server-common/src/test/java/org/a2aproject/sdk/server/events/EventConsumerTest.java
@@ -237,6 +237,41 @@ public class EventConsumerTest {
     }
 
     @Test
+    public void testConsumeTaskMessageBeforeInputRequired() throws Exception {
+        Task workingTask = Task.builder()
+                .id(TASK_ID)
+                .contextId("session-xyz")
+                .status(new TaskStatus(TaskState.TASK_STATE_WORKING))
+                .build();
+        Message message = fromJson(MESSAGE_PAYLOAD, Message.class);
+        TaskStatusUpdateEvent inputRequiredEvent = TaskStatusUpdateEvent.builder()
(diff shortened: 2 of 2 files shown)


kabir (changes requested):
I believe this PR to be wrong for the same reasons as mentioned in https://github.com/a2aproject/a2a-java/pull/1040#pullrequestreview-4905408070

Essentially the agents should either:
- Emit a Message. No Task is created, and is for single reply interactions
- Create a Task for long running tasks

The two are mutually exclusive

dangzitou:
Thank you for the clarification. I agree that a single-reply Message and a Task stream are mutually exclusive in the protocol, so this PR is not the right change. Closing it rather than pursuing the incorrect behavior.

PR #1061, opened 2026-08-13
fix(jsonrpc): prefer a populated content key when decoding a Part

### LINKED ISSUE #1062: [Bug]: Part decoding discards a populated content key when an empty placeholder precedes it
### What happened

`JsonUtil.PartTypeAdapter.read` picks the `oneOf` discriminator as the first content key
whose value is not JSON `null`. An empty string is not `null`, so when a producer emits
every content key of the flattened `oneOf` and leaves the unset ones as `""`, the first
empty placeholder wins and the populated key is discarded.

A `DataPart` sent as `{"text": "", "data": {...}}` decodes as an empty `TextPart`, and the
`data` payload is silently lost — no exception, no warning.

The comment directly above the scan already describes the intended behaviour:

```java
// Find the oneOf discriminator, skipping null/empty values to tolerate formats
// where multiple content keys may be present with only one populated
// (e.g., proto serialization with alwaysPrintFieldsWithNoPresence).
```

Only `null` is skipped; `""` is not. So the exact producer the comment names — proto JSON
with `alwaysPrintFieldsWithNoPresence` — is the one case it fails to tolerate.

### Reproduction

```java
String json = """
    {"text": "", "data": {"turns": [{"role": "user", "text": "hi"}]}}
    """;
Part<?> part = JsonUtil.OBJECT_MAPPER.fromJson(json, Part.class);
// expected: DataPart carrying the turns
// actual:   TextPart("")   — the data payload is gone
```

Observed against `org.a2aproject.sdk:a2a-java-sdk-jsonrpc-common`:

| payload | decodes as |
|---|---|
| `{"text": "", "data": {…}}` | **TextPart("")** — `data` lost |
| `{"data": {…}, "text": ""}` | DataPart (correct) |
| `{"text": null, "data": {…}}` | DataPart (correct) |
| `{"url": "", "data": {…}}` | **FilePart** — `data` lost |

The result depends on JSON key order, which producers are not obliged to control.

### Versions

Reproduces on `1.0.0.Final` and on `1.2.0.Final` and current HEAD.

### Expected

The first content key with a *populated* value should win, so that a producer emitting
empty placeholders round-trips. A part whose only content key is a deliberately empty
string (`{"text": ""}`) must still decode as an empty `TextPart`.

### Impact

An agent receiving such a message answers on input the caller believes it sent. There is
no error to detect, so the failure is only visible as degraded output.

### Relevant log output

```shell

```

### Code of Conduct

rmlopes:
## Summary

- Prefer populated flattened `Part` content keys over empty-string placeholders.
- Preserve deliberately empty text parts via a document-order fallback.
- Reject non-string values for string-backed content keys.
- Add regression coverage for data, URL, raw, null, empty, falsy, and key-order cases.

## Root cause

The discriminator selected the first recognized non-null content key. Some protobuf JSON producers emit every flattened oneOf key and represent unset strings as `""`, causing an empty placeholder to hide the populated payload.

## Validation

- Added and ran focused regression tests in `PartSerializationTest`.

This fixes #1062

### DIFF
--- a/jsonrpc-common/src/main/java/org/a2aproject/sdk/jsonrpc/common/json/JsonUtil.java
+++ b/jsonrpc-common/src/main/java/org/a2aproject/sdk/jsonrpc/common/json/JsonUtil.java
@@ -17,6 +17,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -508,8 +509,8 @@ public class JsonUtil {
      * This adapter handles polymorphic deserialization, creating the
      * appropriate subclass instance (TextPart, FilePart, or DataPart) based on available fields.
      * <p>
-     * The adapter uses a two-pass approach: first reads the JSON as a tree to inspect the "kind"
-     * field, then deserializes to the appropriate concrete type.
+     * The adapter reads the JSON as a tree so that the content key can be inspected before
+     * deserializing to the appropriate concrete type.
      *
      * @see Part
      * @see TextPart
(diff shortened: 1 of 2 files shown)


ehsavoie (approved):

kabir:
Thanks @rmlopes !

PR #1066, opened 2026-08-14
fix: validate A2A version and extensions on all JSON-RPC and REST operations

### LINKED ISSUE #1035: [Bug]: A2A-Version is only validated on message sends, not on the other eight operations
### What happened?

Section 3.6.2 of the specification is a `MUST` on every request:

> Agents **MUST** process requests using the semantics of the requested `A2A-Version` (matching `Major.Minor`). If the version is not supported by the interface, agents **MUST** return a `VersionNotSupportedError`.
>
> Agents **MUST** interpret empty value as 0.3 version.

The SDK calls `A2AVersionValidator.validateProtocolVersion` at exactly two call sites in each transport, both of them message sends:

| Transport | Validated | Not validated |
|---|---|---|
| `transport/jsonrpc` `JSONRPCHandler` | `onMessageSend`, `onMessageSendStream` | `onGetTask`, `onListTasks`, `onCancelTask`, `onSubscribeToTask`, and the four push notification config methods |
| `transport/rest` `RestHandler` | `sendMessage`, `sendStreamingMessage` | same |
| `transport/grpc` `GrpcHandler` | `sendMessage`, `sendStreamingMessage` | same |

So a client sending an `A2A-Version` the interface does not support is correctly refused when it sends a message, and served normally when it calls `GetTask`, `ListTasks`, `CancelTask`, `SubscribeToTask` or any of the push notification config operations. Whether the version is enforced depends on which operation is invoked rather than on the interface, which is the thing 3.6.2 is written to prevent: a client can conclude from a successful `GetTask` that its version is accepted, then have the next `SendMessage` refused.

The good news is that all three bindings behave identically here, so this is one behaviour to change rather than a per binding divergence.

### Suggested fix

Move the check to a place every operation passes through, for instance the request handler decorator chain that already wraps `RequestHandler`, so it cannot be forgotten when an operation is added. A test per binding that calls `GetTask` with an unsupported `A2A-Version` and expects `VersionNotSupportedError` would pin it.

### Reproducer

Added to `JSONRPCHandlerTest` next to the existing version tests, using the same card and the same incompatible context as `testVersionNotSupportedErrorOnMessageSend`, so the only difference is the operation invoked:

```java
@Test
public void testVersionNotSupportedErrorOnGetTask() throws Exception {
    AgentCard agentCard = cardWithProtocolVersion10();
    JSONRPCHandler handler = new JSONRPCHandler(agentCard, requestHandler, internalExecutor);
    taskStore.save(MINIMAL_TASK, false);

    ServerCallContext contextWithVersion = new ServerCallContext(
            UnauthenticatedUser.INSTANCE,
            Map.of("foo", "bar"),
            new HashSet<>(),
            "2.0"); // Incompatible version

    GetTaskRequest request = new GetTaskRequest("1", new TaskQueryParams(MINIMAL_TASK.id()));
    GetTaskResponse response = handler.onGetTask(request, contextWithVersion);

    assertInstanceOf(VersionNotSupportedError.class, response.getError());
    assertNull(response.getResult());
}
```

Run together with the existing message-send test, the existing one passes and this one fails. I am happy to open a PR with the test alone if that is a useful place to start.

### One thing I noticed and did not chase

I first wrote the same test for `onCancelTask`. It reproduces the same gap, but the forked JVM never exits afterwards: the cancel path taps a queue and leaves a consumer thread alive. The pre-existing `testOnCancelTaskSuccess` completes in 0.25s, so something about a freshly constructed handler is involved. I left that test out rather than ship something that hangs the build, but it may be worth a look on your side.

### Relevant log output

```shell
$ mvn -pl transport/jsonrpc test \
    -Dtest='JSONRPCHandlerTest#testVersionNotSupportedErrorOnGetTask+testVersionNotSupportedErrorOnMessageSend'

[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.259 s
        <<< FAILURE! -- in JSONRPCHandlerTest
[ERROR] JSONRPCHandlerTest.testVersionNotSupportedErrorOnGetTask -- Time elapsed: 0.218 s
        <<< FAILURE!
org.opentest4j.AssertionFailedError: Unexpected null value,
        expected: <org.a2aproject.sdk.spec.VersionNotSupportedError> but was: <null>

# the pre-existing testVersionNotSupportedErrorOnMessageSend passes in the same run
```

Call sites, for reference:

```shell
$ git grep -n "validateProtocolVersion" -- '*/src/main/java/*'
server-common/.../version/A2AVersionValidator.java:33:    public static void validateProtocolVersion(
transport/grpc/.../handler/GrpcHandler.java:202:            A2AVersionValidator.validateProtocolVersion(
transport/grpc/.../handler/GrpcHandler.java:398:            A2AVersionValidator.validateProtocolVersion(
transport/jsonrpc/.../handler/JSONRPCHandler.java:233:            A2AVersionValidator.validateProtocolVersion(
transport/jsonrpc/.../handler/JSONRPCHandler.java:290:            A2AVersionValidator.validateProtocolVersion(
transport/rest/.../handler/RestHandler.java:230:            A2AVersionValidator.validateProtocolVersion(
transport/rest/.../handler/RestHandler.java:298:            A2AVersionValidator.validateProtocolVersion(
```

### Environment

- a2a-java `main` at `4f718e12` (`1.2.1.Final-SNAPSHOT`), behaviour identical in `1.2.0.Final`

### Code of Conduct

omatheusmesmo:
# Description

`A2AVersionValidator` was called from two places in each transport, both of them message sends. A client on an unsupported version was refused on `sendMessage` and served normally on the other nine operations. #1042 fixed this for gRPC; this does the same for JSON-RPC and REST.

Both handlers receive the `ServerCallContext` as a parameter rather than building it, so there is no `createCallContext` to centralize in. A private `validateRequest` carries the pair of checks and every operation calls it before reaching the request handler.

## Behaviour change

A client that sends no `A2A-Version` is now refused on every operation rather than only on message sends, since section 3.6.2 makes the empty value mean 0.3 and a 1.0-only card rejects it.

Four REST tests called `GET /tasks/{id}` without the header; they now send it, as section 3.6.1 requires of clients on every request. The SDK's own REST and JSON-RPC clients already send it, as does the TCK.

## Tests

Nine new tests per transport, one for each operation that now validates.

- [ ] Follow the [`CONTRIBUTING` Guide](../CONTRIBUTING.md).
- [ ] Make your Pull Request title in the <https://www.conventionalcommits.org/> specification.
- [ ] Ensure the tests pass
- [ ] Appropriate READMEs were updated (if necessary)

Fixes #1035 🦕

### DIFF
--- a/transport/jsonrpc/src/main/java/org/a2aproject/sdk/transport/jsonrpc/handler/JSONRPCHandler.java
+++ b/transport/jsonrpc/src/main/java/org/a2aproject/sdk/transport/jsonrpc/handler/JSONRPCHandler.java
@@ -233,8 +233,7 @@ public class JSONRPCHandler {
      */
     public SendMessageResponse onMessageSend(SendMessageRequest request, ServerCallContext context) {
         try {
-            A2AVersionValidator.validateProtocolVersion(resolveAgentCard(), context);
-            A2AExtensions.validateRequiredExtensions(resolveAgentCard(), context);
+            validateVersionAndExtensions(context);
             EventKind taskOrMessage = requestHandler.onMessageSend(request.getParams(), context);
             return new SendMessageResponse(request.getId(), taskOrMessage);
         } catch (A2AError e) {
@@ -290,8 +289,7 @@ public class JSONRPCHandler {
         }
 
         try {
-            A2AVersionValidator.validateProtocolVersion(resolveAgentCard(), context);
-            A2AExtensions.validateRequiredExtensions(resolveAgentCard(), context);
+            validateVersionAndExtensions(context);
             Flow.Publisher<StreamingEventKind> publisher =
                     requestHandler.onMessageSendStream(request.getParams(), context);
             // We can't use the convertingProcessor convenience method since that propagates any errors as an error handled
@@ -331,6 +329,7 @@ public class JSONRPCHandler {
      */
     public CancelTaskResponse onCancelTask(CancelTaskRequest request, ServerCallContext context) {
         try {
+            validateVersionAndExtensions(context);
             Task task = requestHandler.onCancelTask(request.getParams(), context);
             if (task != null) {
                 return new CancelTaskResponse(request.getId(), task);
@@ -387,6 +386,7 @@ public class JSONRPCHandler {
         }
         requestHandler.authorizeTaskAccess(request.getParams().id(), context, TaskOperation.SUBSCRIBE_TO_TASK);
         try {
+            validateVersionAndExtensions(context);
             Flow.Publisher<StreamingEventKind> publisher =
                     requestHandler.onSubscribeToTask(request.getParams(), context);
             // We can't use the convertingProcessor convenience method since that propagates any errors as an error handled
@@ -431,6 +431,7 @@ public class JSONRPCHandler {
                     new PushNotificationNotSupportedError());
         }
         try {
+            validateVersionAndExtensions(context);
             TaskPushNotificationConfig config =
                     requestHandler.onGetTaskPushNotificationConfig(request.getParams(), context);
             return new GetTaskPushNotificationConfigResponse(request.getId(), config);
@@ -473,6 +474,7 @@ public class JSONRPCHandler {
                     new PushNotificationNotSupportedError());
         }
         try {
+            validateVersionAndExtensions(context);
             TaskPushNotificationConfig config =
                     requestHandler.onCreateTaskPushNotificationConfig(request.getParams(), context);
             return new CreateTaskPushNotificationConfigResponse(request.getId(), config);
@@ -509,6 +511,7 @@ public class JSONRPCHandler {
      */
     public GetTaskResponse onGetTask(GetTaskRequest request, ServerCallContext context) {
         try {
+            validateVersionAndExtensions(context);
             Task task = requestHandler.onGetTask(request.getParams(), context);
             return new GetTaskResponse(request.getId(), task);
         } catch (A2AError e) {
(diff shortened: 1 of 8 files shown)


transport/jsonrpc/src/main/java/org/a2aproject/sdk/transport/jsonrpc/handler/JSONRPCHandler.java
```diff
@@ -689,6 +696,20 @@ public GetExtendedAgentCardResponse onGetExtendedCardRequest(
...
+    /**
+     * Validates the requested protocol version and extensions against the agent card.
+     *
+     * <p>Section 3.6.2 of the specification requires this on every request, so every operation
+     * calls it before reaching the request handler.
+     *
+     * @param context the server call context carrying the requested version and extensions
+     * @throws A2AError if the requested version or a required extension is not supported
+     */
+    private void validateRequest(ServerCallContext context) throws A2AError {
```
ehsavoie: Maybe use something that better explains what this is doing like: validateVersionAndExtensions

transport/jsonrpc/src/test/java/org/a2aproject/sdk/transport/jsonrpc/handler/JSONRPCHandlerTest.java
```diff
@@ -44,12 +44,15 @@
 import org.a2aproject.sdk.jsonrpc.common.wrappers.SendStreamingMessageRequest;
 import org.a2aproject.sdk.jsonrpc.common.wrappers.SendStreamingMessageResponse;
 import org.a2aproject.sdk.jsonrpc.common.wrappers.SubscribeToTaskRequest;
+import jakarta.enterprise.inject.Instance;
```
ehsavoie: I think the import is not at its 'right' place :)

ehsavoie (changes requested):
Some nit pick changes, overall good :)

omatheusmesmo:
> Some nit pick changes, overall good :)

Thanks for your review, I addressed them!

PR #1067, opened 2026-08-14
feat(jsonrpc): use regex named group for path-tenant routing

### LINKED ISSUE #1034: [Bug]: JSON-RPC server documents tenant-prefixed routing but registers no route that can match it
### What happened?

Context: I found this while evaluating what the SDK can support, for an A2A server extension I am building for Quarkus LangChain4j. It started from a [stalled PR](https://github.com/quarkiverse/quarkus-langchain4j/pull/1994) by another contributor, who is not involved in this work. The goal is to land the extension in quarkus-langchain4j proper.

The JSON-RPC reference server documents tenant-prefixed routing and implements the extraction for it, but never registers a route that can match a tenant-prefixed path, so the documented behaviour is unreachable.

`reference/jsonrpc/.../quarkus/A2AServerRoutes` says in its class javadoc:

> Tenant identification is extracted from the request path:
> - `POST /` → empty tenant
> - `POST /tenant1` → tenant "tenant1"
> - `POST /tenant1/` → tenant "tenant1" (trailing slash stripped)

and `extractTenant(RoutingContext)` implements exactly that. But `setupRoutes` registers only:

```java
router.post("/")
router.get("/.well-known/agent-card.json")
```

In Vert.x a path that does not end in `*` is matched exactly, so `POST /tenant1` never reaches the handler and `extractTenant` can only ever return the empty tenant.

The REST reference implementation does this correctly, with regex routes:

```java
router.postWithRegex("^\\/(?<tenant>[^\\/]*\\/?)message:send$")
```

### Reproduced empirically

A Quarkus application with `a2a-java-sdk-reference-jsonrpc` on the classpath, posting the same JSON-RPC body to each documented path:

| Request | Observed | Documented |
|---|---|---|
| `POST /` | 200 | empty tenant, OK |
| `POST /tenant1` | **405 Method Not Allowed** | tenant `"tenant1"` |
| `POST /tenant1/` | **405 Method Not Allowed** | tenant `"tenant1"` |
| `GET /tenant1/.well-known/agent-card.json` | **404** | n/a |

The 405 rather than 404 comes from another GET-only route matching the path; no POST route matches a tenant-prefixed path at all.

### Why the tests do not catch it

`A2AServerRoutesTest` verifies tenant extraction by mocking the routing context:

```java
when(mockRoutingContext.normalizedPath()).thenReturn("/tenant1");
```

That exercises the extraction in isolation but never goes through the router, so the missing route is invisible. No JSON-RPC test issues a real HTTP request to a tenant-prefixed path.

### Impact

Per [Multi-Tenancy and Multi-Agent Routing](https://a2a-protocol.org/latest/topics/multi-tenancy/), the `tenant` field is one of the three sanctioned routing mechanisms. Today it works over REST but not over JSON-RPC, even though the JSON-RPC implementation documents it as supported, so an agent's routing behaviour depends on which transport a client happens to pick.

### Suggested fix

Register the JSON-RPC endpoint with a tenant-capable route, mirroring what REST already does, and add a test that performs a real HTTP request to a tenant-prefixed path.

The javadoc, `extractTenant` and the two exact-match route registrations are all unchanged from `1.1.0.Final`, where the table above was measured.

### Relevant log output

```shell
$ curl -s -o /dev/null -w '%{http_code}\n' -X POST http://localhost:8080/ \
    -H 'Content-Type: application/json' -H 'A2A-Version: 1.0' -d '{...}'
200

$ curl -s -o /dev/null -w '%{http_code}\n' -X POST http://localhost:8080/tenant1 \
    -H 'Content-Type: application/json' -H 'A2A-Version: 1.0' -d '{...}'
405

$ curl -s -o /dev/null -w '%{http_code}\n' -X POST http://localhost:8080/tenant1/ \
    -H 'Content-Type: application/json' -H 'A2A-Version: 1.0' -d '{...}'
405
```

### Environment

- a2a-java `1.2.0.Final` (measured on `1.1.0.Final`; route registration unchanged)
- Quarkus 3.33.2

### Code of Conduct

ehsavoie:
- Replace fixed POST / route with wildcard regex carrying (?<tenant>...) so multi-segment paths (e.g. /org/team) are captured without string manipulation
- Validate consistency: URL tenant ≠ body tenant → InvalidParamsError
- Add @Priority(1) to test helper routes so Vert.x evaluates them before the A2AServerRoutes catch-all (priority 20)

Fixes #1034 🦕


omatheusmesmo:
Hi @ehsavoie , thanks.

Ran the paths from the issue against this branch and against main, same probes both times.

| Request | main | this PR |
|---|---|---|
| `POST /tenant1/` | 405 | tenant `"tenant1"` |
| `GET /.well-known/agent-card.json` | 200 | 200 |
| `GET /some-unmatched-path` | 404 | 405 |
| `DELETE /some-unmatched-path` | 404 | 405 |

Behaves as expected. Path tenant, multi-segment, trailing slash and the body/URL mismatch validation all work.

One thing worth a look: the catch-all `^/(?<tenant>.*)$` matches every path and is POST only, so any non-POST request to any unmatched path now returns 405 where it returned 404. That reaches applications sharing the HTTP port, not just A2A paths.

ehsavoie:
> One thing worth a look: the catch-all ^/(?<tenant>.*)$ matches every path and is POST only, so any non-POST request to any unmatched path now returns 405 where it returned 404. That reaches applications sharing the HTTP port, not just A2A paths.


@omatheusmesmo Yes but we can't just filter on the tenant. You application would have to have something after the second / Maybe we could make that multitenancy configurable so that the catch all route would only be activated by setting a configuration property? WDYT ?

ehsavoie:
@omatheusmesmo would the changes I've added in the right direction ?

omatheusmesmo (commented):
Checked out the branch and probed it, toggling only the property.

| Request | `multitenancy=false` | PR head (default) |
|---|---|---|
| `POST /` | 200 | 200 |
| `POST /acme` | 404 | 200, tenant `acme` |
| `POST /acme/` | 404 | 404 |
| `POST /org/team` | 404 | 404 |
| `GET /unmapped` (no content type) | 404 | 415 |
| `GET /unmapped` (json content type) | 404 | 200, handler ran |
| `PUT /unmapped` (json content type) | 404 | 200, SendMessage executed |
| `DELETE /unmapped` (no content type) | 404 | 415 |

Three things I would change, then one question.

**1. The multitenancy route is not bound to a method.** `router.routeWithRegex(...)` has no `.method(POST)`, unlike the single tenant branch, so any verb carrying a JSON content type reaches the handler.

**2. `POST /acme/` no longer matches.** `[^/]*` cannot capture a slash, so the trailing slash stripping in `extractTenant` is unreachable and `testTenantExtraction_TrailingSlash` mocks a `pathParam("tenant")` value the router cannot produce. The class javadoc and `multi-tenancy.md` still document that case, and also `POST /org/team`, which the route rejects by design now.

**3. `MultiVersionJSONRPCRoutes` and `A2AServerRoutes_v0_3` ignore the property**, using `^/(?<tenant>.*)$` and `^/.*$`. On `tests/multiversion/jsonrpc` with the flag off, `POST /acme` and `POST /org/team` still return 200, and non-POST verbs on unmapped paths return 405.

I would also default the property to `false`. With `true`, a card with `url: "https://host/rpc"` and `tenant: "billing"` breaks: Section 8.3.2 has the client post to `/rpc` with `params.tenant` set to `billing`, and this branch answers

```
{"error":{"code":-32602,"message":"Tenant mismatch: URL path tenant 'rpc' does not match params tenant 'billing'"}}
```

`false` also makes the PR non breaking.

Question: every request message in `a2a.proto` says `tenant` must match the `AgentInterface` tenant from the card, and nothing checks that. In scope here, or its own issue?

ehsavoie:
@omatheusmesmo After some discussions with the team we agree that this approach, being out of the specs, is not what we should do. I'm going to close it and provide something different based on some internal routing

PR #1071, opened 2026-08-14
feat(http-client): add CDI provider and priority-ordered factory fallback

### LINKED ISSUE #135: [Feat]: Make it possible to plug in other A2AHttpClient implements for the server side

ehsavoie:
- new extras/http-client-cdi module: CdiA2AHttpClientProvider (priority 200) resolves a user-provided A2AHttpClient CDI bean via CDI.current()
- factory iterates providers sorted by descending priority, catching exceptions and falling back to the next provider instead of failing

Fixes #135 🦕

### DIFF
--- a/boms/extras/pom.xml
+++ b/boms/extras/pom.xml
@@ -46,6 +46,11 @@
             </dependency>
             <dependency>
                 <groupId>${project.groupId}</groupId>
+                <artifactId>a2a-java-sdk-http-client-cdi</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>${project.groupId}</groupId>
                 <artifactId>a2a-java-sdk-opentelemetry-common</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/boms/extras/src/it/extras-usage-test/pom.xml
+++ b/boms/extras/src/it/extras-usage-test/pom.xml
@@ -54,6 +54,10 @@
         </dependency>
         <dependency>
             <groupId>org.a2aproject.sdk</groupId>
+            <artifactId>a2a-java-sdk-http-client-cdi</artifactId>
+        </dependency>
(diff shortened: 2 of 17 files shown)


malladinagarjuna2:
one thing worth  considering, when the CDI lookup is ambiguous rather than absent, the fallback is completely silent. Instance.isResolvable() returns false for both "no bean" and "two beans", so a user who accidentally has two A2AHttpClient beans silently gets JdkA2AHttpClient instead of losing whatever custom TLS they added the module for.

It's invisible because the failure is logged at FINE, below JUL's default INFO threshold. Same test run, only the log level differs:

At default log level, nothing is printed at all. Running the exact same test with `.level=FINE` surfaces this:

```
FINE: Provider cdi skipped
java.lang.IllegalStateException: No unambiguous A2AHttpClient CDI bean found
    at CdiA2AHttpClientProvider.create(CdiA2AHttpClientProvider.java:42)
    at A2AHttpClientFactory.lambda$create$1(A2AHttpClientFactory.java:78)
```

"No provider available" is a normal condition and worth keeping at `FINE`. But a provider that advertised priority 200 and then failed usually means the user *intended* to use it — that case might deserve `WARNING`.

Separately: the current tests mock `CDI.current()` statically, so they don't exercise real CDI resolution. I have three Weld SE tests locally , real-container resolution, the end-to-end push-notification path, and the ambiguous-bean case.

ehsavoie:
If you have such a test feel free to send a pr on my branch.
I agree with the WARNING level

PR #1084, opened 2026-08-21
feat(server)!: add multitenancy support with per-tenant routing

### LINKED ISSUE #1034: [Bug]: JSON-RPC server documents tenant-prefixed routing but registers no route that can match it

ehsavoie:
- Add AgentExecutorRouter and AgentCardRouter SPI in server-common
- Add @Tenant CDI qualifier and CDI router implementations in extras/multitenancy
- Wire tenant-aware resolution in DefaultRequestHandler and all transport handlers
- Add tenant field to RequestContext for execute/cancel flows
- Update GetExtendedAgentCardRequest to carry GetExtendedAgentCardParams

Fixes #1034 🦕


ehsavoie:
@omatheusmesmo a new approach that makes more sense I think

server-common/src/main/java/org/a2aproject/sdk/server/util/CdiUtils.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public static <T> @Nullable T getIfResolvable(@Nullable Instance<T> instance) {
+        if (instance != null && instance.isResolvable()) {
+            return instance.get();
+        }
+        return null;
+    }
+
+    /**
+     * Finds the single default (non-tenant) bean of a given type, optionally requiring
```
kabir: Does 'non-tenant' make sense here? Seems to tie it to the tenant stuff rather than a generic method

server-common/src/main/java/org/a2aproject/sdk/server/util/CdiUtils.java
```diff
@@ -0,0 +1,83 @@
...
+     * a specific qualifier. Warns on duplicate defaults and validates that the bean scope
+     * is {@code @ApplicationScoped} or {@code @Singleton}.
+     *
+     * @param allBeans          the CDI instance containing all beans of type {@code T}
+     * @param excludeQualifier  qualifier that marks tenant-specific beans (skipped during resolution)
```
kabir: tenant is used again

transport/grpc/src/main/java/org/a2aproject/sdk/transport/grpc/handler/GrpcHandler.java
```diff
@@ -563,12 +565,19 @@ private void completeStream() {
     public void getExtendedAgentCard(org.a2aproject.sdk.grpc.GetExtendedAgentCardRequest request,
                            StreamObserver<org.a2aproject.sdk.grpc.AgentCard> responseObserver) {
         try {
-            if (!getAgentCard().capabilities().extendedAgentCard()) {
+            if (!resolveAgentCard().capabilities().extendedAgentCard()) {
                 handleError(responseObserver, new UnsupportedOperationError());
                 return;
             }
-            ServerCallContext context = createCallContext(responseObserver);
```
kabir, line 570: Removing this 2A causes protocol version validation and required extensions validation to no longer run for gRPC getExtendedAgentCard requests. This violates Section 3.6.2 of the A2A spec which requires version checking on every request.

kabir (changes requested):
Overall looks good, some minor stuff.

But should we put this into compat-0.3 too?

docs/content/dev/extras.md
```diff
@@ -40,6 +40,10 @@ JPA-backed replacements for the in-memory stores, providing database persistence
 
 Replaces the default `InMemoryQueueManager` with event replication across multiple A2A server instances via message brokers. Required for multi-instance deployments. The core module (`a2a-java-queue-manager-replicated-core`) pairs with a MicroProfile Reactive Messaging strategy (`a2a-java-queue-manager-replication-mp-reactive`) supporting Apache Kafka, Pulsar, or AMQP. You can also write your own `ReplicationStrategy`.
 
+## [Multi-Tenancy](multi-tenancy)
```
kabir: This should link to `../extra/multi-tenancy`.

However, there is no such file for the extra implementation

kabir (changes requested):
@ehsavoie Do we need an extra/multi-tenancy file?

docs/content/dev/extras.md
```diff
@@ -40,6 +40,10 @@ JPA-backed replacements for the in-memory stores, providing database persistence
 
 Replaces the default `InMemoryQueueManager` with event replication across multiple A2A server instances via message brokers. Required for multi-instance deployments. The core module (`a2a-java-queue-manager-replicated-core`) pairs with a MicroProfile Reactive Messaging strategy (`a2a-java-queue-manager-replication-mp-reactive`) supporting Apache Kafka, Pulsar, or AMQP. You can also write your own `ReplicationStrategy`.
 
+## [Multi-Tenancy](../extra/multi-tenancy)
```
kabir, line 43: We need to create the extra/multi-tenancy.md file

docs/content/dev/multi-tenancy.md
```diff
@@ -0,0 +1,135 @@
...
+title: Multi-Tenancy
+description: Serve multiple tenants from a single A2A server — per-tenant AgentExecutor and AgentCard routing with CDI qualifiers.
+layout: page
+---
+
+# Multi-Tenancy
+
+Multi-tenancy lets a single A2A server provide different agent behavior per tenant. Each tenant can have its own `AgentExecutor` (business logic) and extended `AgentCard` (capabilities, skills, metadata). Requests without a tenant — or with an unknown tenant — fall back to the default beans automatically.
+
+## Setup
```
kabir: I think probably everything here to ~ L121 should go in the missing extra/multi-tenancy.md file.

Maybe this section can explain the overall concepts (in case people want to write their own), and then the extra/multi-tenancy.md one explains how our implementation works?

kabir (commented):
Missing extra/multi-tenancy.md

kabir (approved):

PR #1092, opened 2026-08-25
fix: allow omitted task push config id

### LINKED ISSUE #1081: [Bug]: TaskPushNotificationConfig.id is enforced as required, contradicting the spec where id is optional on create
## Summary

Per the A2A specification (Section 3.1.7, "Create Push Notification Config"), the `id` field of a push notification config is **optional** when creating one — the only required input is `url`, and the server assigns an ID if none is provided. The runtime request paths honor this, but the Java domain record `TaskPushNotificationConfig` enforces `id` as non-null, which contradicts the spec, the gRPC proto definition, and the store interface's own documented contract.

## Spec reference

- [A2A Protocol Spec — Create Push Notification Config](https://a2a-protocol.org/latest/specification/): `id` is **not** required on create; the server generates and returns the assigned ID.

### Consistent with the spec

- `spec-grpc/src/main/proto/a2a.proto` (`message TaskPushNotificationConfig`, lines 466–480): `string id = 2;` is **not** marked `(google.api.field_behavior) = REQUIRED`; only `url` is.
- All request entry points (JSON-RPC `JSONRPCUtils.java:224-230`, REST `RestHandler.java:376-385`) parse the body into a **proto** builder first, so a missing `id` arrives as the protobuf default empty string `""`, and `InMemoryPushNotificationConfigStore.setInfo` (`:51`) defaults an empty id to the task ID (`if (notificationConfig.id().isEmpty()) builder.id(taskId)`). So a client omitting `id` does get a config back with an assigned ID.

### Inconsistent with the spec

- `spec/src/main/java/org/a2aproject/sdk/spec/TaskPushNotificationConfig.java`:
  - Compact constructor (line 48): `Assert.checkNotNullParam("id", id);`
  - `Builder.build()` (line 177): `Assert.checkNotNullParam("id", id);`

This makes `id` **required** at the domain-API level, so:

1. `TaskPushNotificationConfig.builder().url(...).build()` throws `IllegalArgumentException` when `id` is omitted — but the spec says `id` is optional and the server assigns it.
2. `PushNotificationConfigStore.setInfo` documents "If `notificationConfig.id()` is null or empty, it's set to the task ID" (`PushNotificationConfigStore.java:91`), but the **null branch is dead code** — a null `id` can never be constructed through the public API.
3. `InMemoryPushNotificationConfigStore.setInfo:51` calls `notificationConfig.id().isEmpty()`; if a null `id` ever reached it (as the Javadoc promises), it would NPE instead of defaulting.

The current "optional id works at runtime" behavior is load-bearing on the fact that the mapper `TaskPushNotificationConfigMapper.fromProto` does **not** apply `emptyToNull` to `id` (it does for `taskId`/`token`/`tenant`) — it only survives because proto yields `""` rather than `null`.

## Related (secondary) inconsistency

`GetTaskPushNotificationConfigParams.java:31-32`: the Javadoc describes `id` as "optional specific configuration ID to retrieve", but the constructor asserts `checkNotNullParam("id", id)` — so the default config cannot be retrieved without specifying an id.

## Expected behavior

- `id` should be optional when constructing/creating a `TaskPushNotificationConfig`; the server assigns an ID (e.g., defaults to the task ID) when it is absent, and returns the config with the assigned ID.
- Constructing `TaskPushNotificationConfig.builder().url(...).build()` should succeed, mirroring the spec and the 0.3 compat record `PushNotificationConfig_v0_3` (which only validates `url`).

### Relevant log output

```shell

```

### Code of Conduct

hutiefang76:
## Description

Allow `TaskPushNotificationConfig` to omit its `id` when a client creates a push notification configuration, matching the A2A specification. The in-memory store now assigns the task ID when the supplied ID is null or empty, preserving its existing defaulting behavior.

## Tests

- `mvn -pl spec,server-common -am -Dtest=TaskPushNotificationConfigTest,InMemoryPushNotificationConfigStoreTest -Dsurefire.failIfNoSpecifiedTests=false test`
  - `TaskPushNotificationConfigTest`: 1 passed
  - `InMemoryPushNotificationConfigStoreTest`: 33 passed

- [x] Follow the CONTRIBUTING guide
- [x] Use a conventional commit title
- [x] Tests pass
- [x] README changes are not needed for this API-contract correction

Fixes #1081 🦕

### DIFF
--- a/server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStore.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStore.java
@@ -48,11 +48,13 @@ public class InMemoryPushNotificationConfigStore implements PushNotificationConf
     public TaskPushNotificationConfig setInfo(TaskPushNotificationConfig notificationConfig) {
         String taskId = Assert.checkNotNullParam("taskId", notificationConfig.taskId());
         TaskPushNotificationConfig.Builder builder = TaskPushNotificationConfig.builder(notificationConfig);
-        if (notificationConfig.id().isEmpty()) {
+        String requestedConfigId = notificationConfig.id();
+        boolean configIdIsMissing = requestedConfigId == null || requestedConfigId.isEmpty();
+        String configId = configIdIsMissing ? taskId : requestedConfigId;
+        if (configIdIsMissing) {
             builder.id(taskId);
         }
         TaskPushNotificationConfig config = builder.build();
-        String configId = config.id();
         int maxPerTask = PushNotificationConfigStore.maxPushConfigsPerTask(configProvider);
 
         pushNotificationInfos.compute(taskId, (key, list) -> {
--- a/server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStore.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStore.java
@@ -46,6 +46,12 @@ public class InMemoryPushNotificationConfigStore implements PushNotificationConf
 
     @Override
     public TaskPushNotificationConfig setInfo(TaskPushNotificationConfig notificationConfig) {
+        return setInfo(notificationConfig, false);
+    }
+
+    private TaskPushNotificationConfig setInfo(
+            TaskPushNotificationConfig notificationConfig,
+            boolean allowDefaultConfigUpdate) {
         String taskId = Assert.checkNotNullParam("taskId", notificationConfig.taskId());
         TaskPushNotificationConfig.Builder builder = TaskPushNotificationConfig.builder(notificationConfig);
         String requestedConfigId = notificationConfig.id();
@@ -61,7 +67,7 @@ public class InMemoryPushNotificationConfigStore implements PushNotificationConf
             List<TaskPushNotificationConfig> mutable = list == null ? new ArrayList<>() : new ArrayList<>(list);
             boolean defaultConfigAlreadyExists = configIdIsMissing
                     && mutable.stream().anyMatch(existing -> existing.id() != null && existing.id().equals(configId));
-            if (defaultConfigAlreadyExists) {
+            if (defaultConfigAlreadyExists && !allowDefaultConfigUpdate) {
                 throw new InvalidParamsError("A push notification config with the default ID already exists for task "
                         + taskId + "; specify the config ID explicitly to update it");
             }
@@ -79,8 +85,10 @@ public class InMemoryPushNotificationConfigStore implements PushNotificationConf
 
     @Override
     public TaskPushNotificationConfig setInfo(TaskPushNotificationConfig config, @Nullable String protocolVersion) {
-        TaskPushNotificationConfig result = setInfo(config);
-        protocolVersions.put(result.taskId() + ":" + result.id(), PushNotificationConfigStore.resolveProtocolVersion(protocolVersion));
+        TaskPushNotificationConfig result = setInfo(config, "0.3".equals(protocolVersion));
+        protocolVersions.put(
+                result.taskId() + ":" + result.id(),
+                PushNotificationConfigStore.resolveProtocolVersion(protocolVersion));
         return result;
     }
 
--- a/server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStore.java
+++ b/server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStore.java
@@ -59,6 +59,12 @@ public class InMemoryPushNotificationConfigStore implements PushNotificationConf
 
         pushNotificationInfos.compute(taskId, (key, list) -> {
             List<TaskPushNotificationConfig> mutable = list == null ? new ArrayList<>() : new ArrayList<>(list);
+            boolean defaultConfigAlreadyExists = configIdIsMissing
+                    && mutable.stream().anyMatch(existing -> existing.id() != null && existing.id().equals(configId));
+            if (defaultConfigAlreadyExists) {
+                throw new InvalidParamsError("A push notification config with the default ID already exists for task "
+                        + taskId + "; specify the config ID explicitly to update it");
+            }
             boolean isExistingConfig = mutable.removeIf(
                     existing -> existing.id() != null && existing.id().equals(configId));
             if (!isExistingConfig && mutable.size() >= maxPerTask) {
--- a/server-common/src/test/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStoreTest.java
+++ b/server-common/src/test/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStoreTest.java
@@ -154,7 +154,6 @@ class InMemoryPushNotificationConfigStoreTest {
     public void testSetInfoWithoutConfigId() {
         String taskId = "task1";
         TaskPushNotificationConfig initialConfig = TaskPushNotificationConfig.builder()
-                .id("") // No ID set
                 .url("http://initial.url/callback")
                 .taskId(taskId)
                 .build();
@@ -167,7 +166,6 @@ class InMemoryPushNotificationConfigStoreTest {
         assertEquals(taskId, configResult.configs().get(0).id());
 
         TaskPushNotificationConfig updatedConfig = TaskPushNotificationConfig.builder()
-                .id("") // No ID set
                 .url("http://initial.url/callback_new")
                 .taskId(taskId)
                 .build();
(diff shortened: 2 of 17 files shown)


server-common/src/test/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStoreTest.java
```diff
@@ -152,7 +152,6 @@ public void testSetInfoAppendsToExistingConfig() {
     public void testSetInfoWithoutConfigId() {
         String taskId = "task1";
         TaskPushNotificationConfig initialConfig = TaskPushNotificationConfig.builder()
-                .id("") // No ID set
```
kabir, line 157: Please create a new test rather than changing this existing one. 

Removing `.id("")` repurposes this test from the empty-string path to the null path rather than covering both. 
The empty-string case is the one production actually hits — the mapper doesn't apply emptyToNull to id, so at runtime a missing id arrives as "", never null. 
Recommend keeping the "" case here and adding a separate null-id test (or parameterizing over both), so the load-bearing branch stays asserted.

server-common/src/test/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStoreTest.java
```diff
@@ -165,7 +164,6 @@ public void testSetInfoWithoutConfigId() {
         assertEquals(taskId, configResult.configs().get(0).id());
 
         TaskPushNotificationConfig updatedConfig = TaskPushNotificationConfig.builder()
-                .id("") // No ID set
```
kabir, line 170: Please create a new test rather than changing this existing one.

server-common/src/main/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStore.java
```diff
@@ -48,11 +48,13 @@ public InMemoryPushNotificationConfigStore() {
     public TaskPushNotificationConfig setInfo(TaskPushNotificationConfig notificationConfig) {
         String taskId = Assert.checkNotNullParam("taskId", notificationConfig.taskId());
         TaskPushNotificationConfig.Builder builder = TaskPushNotificationConfig.builder(notificationConfig);
-        if (notificationConfig.id().isEmpty()) {
```
kabir, line 51: Not a comment about this code, but `JpaDatabasePushNotificationConfigStore` has the same latent NPE, and needs fixing too.

kabir (changes requested):
Hi @hutiefang76 - thank you for the PR 👍 

I found a few small issues in the code, but we also need to take `JpaDatabasePushNotificationConfigStore` into account and make some changes, and test this inline with what you have done here. `JpaPushNotificationConfigStoreTest` might be the right place.

Also `GetTaskPushNotificationConfigParams` asserts `id` non-null despite its Javadoc calling it optional. Might as well fix this since it is related.

hutiefang76:
Updated in 5bd4916.

- `JpaDatabasePushNotificationConfigStore` now normalizes a null or empty config ID to the task ID, with a Quarkus/JPA regression test.
- `GetTaskPushNotificationConfigParams` now permits an omitted ID through both its constructor and builder, consistent with its Javadoc. The request handler resolves an omitted ID to the task ID, matching creation and v0.3 compatibility behavior.
- Added focused model and request-handler coverage.

Verified on JDK 17 with:
`mvn -pl spec,server-common,extras/push-notification-config-store-database-jpa -am -Dtest=GetTaskPushNotificationConfigParamsTest,DefaultRequestHandlerTest,JpaPushNotificationConfigStoreTest -Dsurefire.failIfNoSpecifiedTests=false test`

Result: 2 spec tests, 23 request-handler tests, and 16 JPA-store tests passed (one pre-existing disabled test).`

server-common/src/test/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStoreTest.java
kabir: The previous comment should still be addressed.

server-common/src/test/java/org/a2aproject/sdk/server/tasks/InMemoryPushNotificationConfigStoreTest.java
kabir: The previous comment should still be addressed.

kabir (changes requested):

hutiefang76:
Addressed in dab6633.

- Restored the existing empty-id coverage with `.id("")`, so it continues to test the exact value emitted by the mapper.
- Added a separate null-id regression test; it verifies the store derives the config id from the task id when the id is absent.
- The equivalent JPA null/empty normalization is already included in 5bd4916.

Validated with `GetTaskPushNotificationConfigParamsTest`, `DefaultRequestHandlerTest`, `InMemoryPushNotificationConfigStoreTest`, and `JpaPushNotificationConfigStoreTest` (75 tests, 0 failures/errors).`

kabir:
Sorry @hutiefang76 we're doing a lot of merging today, and there is now a conflict :-)

hutiefang76:
Rebased onto current `main` to resolve the merge conflict (latest head: `e4a3d37`). The resolution preserves upstream tenant validation while retaining the optional config-id contract and its empty-id/null-id coverage.\n\nValidated after the rebase with the focused spec, request-handler, in-memory-store, and JPA-store suites: 84 tests, 0 failures/errors; `git diff --check` also passes.

ehsavoie (changes requested):
This is mostly good and correct but I have a behaviour question:
When creating multiple push notification configs for the same task with omitted IDs. Both would default to taskId, so the second setInfo call would silently update the first config. This behaviour should be documented and covered a test. Also I think that at least we should 'trace' that somehow or maybe we should fail. Using empty id only to create.

hutiefang76:
Addressed the remaining review concern in 1c21dcd4.\n\nA missing or empty ID still creates the default config with the task ID, preserving the existing default-get behavior. A second create with an omitted ID now fails with InvalidParamsError instead of silently replacing that default config; clients must send the task ID explicitly to update it. This is implemented consistently in the in-memory and JPA stores, and the public store contract now documents the rule.\n\nAdded coverage for both empty-string and null IDs in both stores. Verified with:\n- mvn -pl server-common -am -Dtest=InMemoryPushNotificationConfigStoreTest -Dsurefire.failIfNoSpecifiedTests=false test (34 passing)\n- mvn -pl extras/push-notification-config-store-database-jpa -am -Dtest=JpaPushNotificationConfigStoreTest -Dsurefire.failIfNoSpecifiedTests=false test (16 passing, 1 skipped)

hutiefang76:
Addressed the v0.3 compatibility regression in `bd875377`.

The duplicate omitted-ID guard added in the previous commit is still enforced for the current multi-config protocol, but v0.3 now preserves its historical single-config set/update behavior. The protocol-aware path is applied consistently to the in-memory and JPA stores, and the v0.3 test adapter now records the correct protocol version.

Added regression coverage for both stores and verified on JDK 21 with:

- `InMemoryPushNotificationConfigStoreTest`, `DefaultRequestHandlerTest`, `JpaPushNotificationConfigStoreTest`, and `GetTaskPushNotificationConfigParamsTest`: 86 tests, 0 failures/errors (1 skipped)
- v0.3 JSON-RPC, gRPC, and REST reference suites: 315 tests, 0 failures/errors (38 skipped)
- `git diff --check`

This fixes the Java 21 CI failures where repeated v0.3 `setPushNotification` calls were rejected instead of updating the single stored config.

hutiefang76:
I traced the four build failures to the multiversion test setup path rather than the production v0.3 route.

The v0.3 client/server path already carries `requestedProtocolVersion = "0.3"`, but `AbstractA2AServerServerTest_v0_3` seeds push configs through the shared `/test/task/{taskId}` endpoint. That test endpoint discarded the protocol version, so the setup call used v1.0 create-only semantics and the second omitted-ID config failed (or leaked an extra config into later assertions).

Commit `05422c57` now:
- sends the standard `A2A-Version: 0.3` header from the v0.3 test helper;
- forwards that header through the JSON-RPC, REST, and gRPC multiversion test endpoints;
- keeps the existing no-header helper behavior for v1.0 tests.

Local JDK 21 verification:
- `MultiVersionJSONRPC_v0_3_Test`: 38 tests, 0 failures/errors
- `MultiVersion_v0_3_RestTest`: 38 tests, 0 failures/errors, 9 skipped
- `MultiVersion_v0_3_GrpcTest`: 38 tests, 0 failures/errors, 9 skipped

Total: 114 tests, 0 failures/errors.

hutiefang76:
@kabir @ehsavoie The latest head `05422c57` addresses the requested store coverage, omitted-ID duplicate behavior, and v0.3 compatibility setup. All 13 GitHub checks are now green, including multiversion JSON-RPC, REST, gRPC, TCK, Windows, and JDK 17/21/25 builds. Could you please take a fresh look when convenient?

ehsavoie (approved):

ehsavoie:
Thanks @hutiefang76

PR #1111, opened 2026-08-31
fix: Multitenancy support for public agent card doesn't match the client.

ehsavoie:
On top of https://github.com/a2aproject/a2a-java/pull/1110
Fixes #1107 🦕

PR #1112, opened 2026-08-31
feat(multitenancy): return 404 for unknown tenant on public agent card

### LINKED ISSUE #1108: [Bug]: When a tenant doesn't exist, we should return a 404
### What happened?

Instead of falling back to the default agent we should return a 404 when the tenant specified doesn't exist. The default tenant should be for non specified tenant (null or empty string).

### Relevant log output

```shell

```

### Code of Conduct

ehsavoie:
On top of https://github.com/a2aproject/a2a-java/pull/1111
Fixes https://github.com/a2aproject/a2a-java/issues/1108 🦕


kabir:
@ehsavoie Needs a rebase

ehsavoie:
@kabir rebased

PR #1126, opened 2026-09-03
chore: Update project URL in pom.xml

jmesnil:


jmesnil:
@kabir I think it needs to be fixed to help setup correctly our Maven Central limits...

PR #1149, opened 2026-09-09
feat(sse): add SSEParserConfig, raise default per-line limit to 1 MB

### LINKED ISSUE #1123: [Bug]: SSE client rejects any event whose data arrives on a single line over 64KB, with no way to configure the limit

ehsavoie:
- Introduce SSEParserConfig record with builder to make SSE parser
  limits (maxLineLength, maxBufferLines, maxBufferChars) configurable
- Raise default per-line limit from 64 KB to 1 MB so large JSON-RPC
  responses are no longer silently dropped
- Thread SSEParserConfig through all HTTP client implementations
  (JDK, Android, OkHttp) via new constructors and provider methods
- Prevent corrupt/skipped event blocks from advancing lastEventId,
  which could cause event loss on reconnect
- Add AbstractA2AHttpClientSSETest for shared cross-client SSE tests
- Document SSEParserConfig usage and security considerations

Fixes #1123 🦕


server-common/src/main/resources/META-INF/a2a-defaults.properties
```diff
@@ -52,6 +52,14 @@ a2a.authorization.required=true
 # CreateTaskPushNotificationConfig calls return UnsupportedOperationError.
 a2a.push-notification.enabled=true
 
+# SSE Parser - Configurable limits for Server-Sent Event parsing (client-side)
+# Maximum characters per raw SSE line (0 = disabled)
+a2a.sse.max-line-length=1048576
```
kabir: These should be documented in the site

kabir (approved):

PR #1156, opened 2026-09-16
fix: preserve push notification order per task

### LINKED ISSUE #775: [Bug]: pushnotification is not support order-preserving of update event

serhiy-bzhezytskyy:
# Description

- [x] Follow the `CONTRIBUTING` Guide.
- [x] Make your Pull Request title in the https://www.conventionalcommits.org/ specification.
- [x] Ensure the tests pass

`sendPushNotification` submitted each push independently to the shared `ForkJoinPool`, so delivery order for the same task depended on scheduling rather than production order — a client could see `working → completed` arrive as `completed → working`. Chains pushes onto a per-task tail future instead: each one only starts after the previous one for the same task finishes, while different tasks stay fully independent (no dedicated thread per task, no manual queue to synchronize).

Added `MainEventBusProcessorPushNotificationOrderTest`: one test proving same-task ordering under a delay pattern that would expose the old race, one proving a slow task never blocks a fast one. Confirmed red-then-green: both fail against the previous implementation (scrambled order), pass against this one. Full `server-common` suite: 519 tests, 0 failures.

There was an earlier attempt at this in #776 (per-task bounded queue + dedicated consumer thread), closed for inactivity after Gemini flagged a blocking `join()` on the critical processing thread and a race in the queue's overflow handling. This avoids both by not using dedicated threads at all.

Fixes #775 🦕


server-common/src/main/java/org/a2aproject/sdk/server/events/MainEventBusProcessor.java
```diff
@@ -393,12 +413,14 @@ private void sendPushNotification(String taskId, StreamingEventKind event, @Null
...
-            pushNotificationExecutor.execute(pushTask);
-        } else {
-            CompletableFuture.runAsync(pushTask);
-        }
+        Executor executor = pushNotificationExecutor != null ? pushNotificationExecutor : ForkJoinPool.commonPool();
+
+        pushNotificationChains.compute(taskId, (id, previous) -> {
+            CompletableFuture<Void> previousOrDone = previous != null ? previous : CompletableFuture.completedFuture(null);
+            CompletableFuture<Void> next = previousOrDone.thenRunAsync(pushTask, executor);
+            next.whenComplete((v, t) -> pushNotificationChains.remove(id, next));
```
kabir: **P2:** When `pushNotificationExecutor` is `Runnable::run`, `thenRunAsync` completes synchronously. The `whenComplete` callback therefore removes `next` during `compute`, after which `compute` installs the completed future back into the map. This leaves one completed entry per task, so the map can grow indefinitely and contradicts the cleanup guarantee.\n\nCould the completion callback be registered after `compute` returns?\n\n```java\nCompletableFuture<Void> next = pushNotificationChains.compute(...);\nnext.whenComplete((v, t) -> pushNotificationChains.remove(taskId, next));\n```

serhiy-bzhezytskyy: Fixed in 8515d095, along the lines you suggested: `whenComplete` is now attached after `compute()` returns, so the map always holds `next` before the callback can try to remove it.

kabir: Thank you :-)

kabir (approved):

server-common/src/main/java/org/a2aproject/sdk/server/events/MainEventBusProcessor.java
serhiy-bzhezytskyy: Thank you too

ehsavoie (approved):

PR #1162, opened 2026-09-20
fix(http-client-vertx): resolve relative redirect Location against the request URI (#1142)

### LINKED ISSUE #1142: [Bug]: Vert.x opt-in redirects do not support relative Location values.
### What happened?

VertxA2AHttpClient.java:576-579 passes Location directly to getAbs()/
postAbs(). HTTP redirects commonly use values such as /collect; these will fail instead of being resolved against the original request URI. Resolve
relative locations before constructing the redirected request.

### Relevant log output

```shell

```

### Code of Conduct

chopmob-cloud:
## Summary

Fixes #1142. The Vert.x client's opt-in redirect following passed the redirect `Location` straight to `webClient.getAbs(location)` (301/302/303) and `postAbs(location)` (307/308), so a relative `Location` (e.g. `/collect`, permitted by RFC 9110/7231) was never resolved against the original request URI and the redirect failed with `MalformedURLException: no protocol`.

## Fix

Resolve the `Location` against the original request URI per RFC 3986 (`URI.create(url).resolve(location)`) in both redirect branches before constructing the redirected request. Absolute Locations pass through unchanged; a malformed base/location falls back to the raw value (prior behavior). Only `VertxPostBuilder.post()` follows redirects (the GET/SSE paths are single-shot by design), so the fix is scoped to those two branches, and no existing redirect-target validation is added or weakened.

## Tests

Added `synchronousPostFollowsRelativeRedirectLocation` (MockServer, single origin): a `302` with `Location: /collect` is now resolved against the base and followed through to the target. The inherited absolute-Location test confirms absolute Locations still work unchanged. Reverting the resolve step reproduces the original `MalformedURLException: no protocol: /collect`. `mvn -pl extras/http-client-vertx -am test` passes (45 tests, 0 failures).


extras/http-client-vertx/src/main/java/org/a2aproject/sdk/client/http/vertx/VertxA2AHttpClient.java
```diff
@@ -533,6 +534,27 @@ public CompletableFuture<Void> getAsyncSSE(
...
+     * reference such as {@code /collect}; RFC 3986 5 reference resolution turns it into the
+     * effective target URI. An already-absolute {@code Location} is returned unchanged. A
+     * malformed base or location falls back to the raw value so any existing handling still
+     * applies.
+     *
+     * @param requestUri the original (absolute) request URI the redirect responded to
+     * @param location the raw {@code Location} header value
+     * @return the absolute redirect target
+     */
+    private static String resolveRedirectLocation(String requestUri, String location) {
```
ehsavoie: This should be package protected so we can add a bunch of unit tests to validate the expected behaviour instead of 'just' a single 'integration' test with mocks

ehsavoie (changes requested):

kabir (approved):

ehsavoie (approved):

ehsavoie:
Thanks @chopmob-cloud

Issue #1012, opened 2026-08-05
fix: handle blank push notification config id consistently across transports


malladinagarjuna2:
`GetTaskPushNotificationConfigParams` rejects a null `id` but permits an
empty string, and the two transports diverge on that input:

- `GrpcTransport` sends `.setId("")`.
- `RestTransport` treats it as absent and requests the trailing-slash
  "default config" URL.

Identical caller code against the same `ClientTransport` method therefore
produces two different requests. This aligns both on <chosen behaviour>
and adds coverage for the blank-id path, which was previously untested.

Also corrects the Javadoc on `GetTaskPushNotificationConfigParams`, which
described `id` as optional and documented `@throws` for `taskId or tenant`
when the constructor actually validates `taskId` and `id`.

PR #1163, opened 2026-09-20
fix(grpc): allow omitted push config ids

56wj:
# fix(grpc): support omitted push notification config ids

## Summary

Follow up on #1092: `GetTaskPushNotificationConfigParams` permits an omitted configuration ID, and `DefaultRequestHandler` resolves null/empty IDs to the task ID. However, `GrpcTransport` still rejects an omitted ID before sending the RPC, unlike `RestTransport`.

- Remove the stale gRPC-only non-null assertion on the optional configuration ID.
- Encode a null configuration ID as the protobuf default empty string; preserve explicit IDs and tenant resolution.
- Add five in-process gRPC tests for constructor/builder omission, empty ID, explicit ID, and tenant precedence.
- Add REST tests asserting that null/empty IDs use the default-configuration GET URL rather than LIST.
- Add `grpc-inprocess` only as a test-scoped dependency. No production dependencies or public API changes.

## Verification

Before the fix, `GrpcTransportTest` ran 5 tests with 3 errors (`IllegalArgumentException: Parameter 'id' may not be null`), all on omitted-ID paths.

After the fix:

```sh
mvn -pl client/transport/grpc,client/transport/rest -am test
```

557 tests across the two transports and their dependency modules passed, with 0 failures, errors, or skipped tests. This is the selected reactor, not the entire repository.

- [x] Follow the CONTRIBUTING guide
- [x] Conventional commit title
- [x] Tests pass
- [x] No README change needed: this restores the existing documented optional-ID contract

Refs #1012; follow-up to #1092. The original issue's null-ID premise predates #1092; this PR addresses the remaining client-side omission rather than choosing a new blank-ID policy.

Developed with AI coding assistance, with failing-before/passing-after regression evidence.


kabir:
Thank you @56wj

